PR #1, opened 2022-05-31
Add scale worker nodes common function

pierDipi:
A common use case across midstream repos is to scale up the worker
nodes.

The way it has been done so far is to always set whatever is the
first machineset to a specified number of workers. Obviously, this
isn't ideal since the sum of worker nodes across the (usual)
multiple machineset might already meet the expected number of
replicas. [1]

In this PR, we are considering every machineset and scale
up only when there are missing nodes.

https://github.com/openshift/knative-eventing/blob/21b347723056d07a96e79ab44744d6f4b8217c26/openshift/e2e-common.sh#L34-L66

### DIFF
--- a/lib.sh
+++ b/lib.sh
@@ -1,0 +1,26 @@
+# scale up worker nodes to at least the provided number of nodes
+# Inputs:
+#  1: number of minimum worker nodes
+function scale_worker_nodes() {
+  echo "Reconcile workers to at least ${1} nodes"
+
+  additional_replicas=$(oc get machineset -n openshift-machine-api | awk '{print $2}' | tail -n +2 | awk -v workers="$1" '{sum+=$1} END {print workers-sum}')
+  echo "Additional replicas ${additional_replicas}"
+
+  if [[ ${additional_replicas} -gt 0 ]]; then
+    machineset="$(oc get machineset -n openshift-machine-api | awk '{print $1}' | tail -n +2 | head -1)"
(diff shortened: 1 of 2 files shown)


aliok:
Can we have the new code written in Python?

lib.sh
```diff
@@ -0,0 +1,26 @@
+# scale up worker nodes to at least the provided number of nodes
+# Inputs:
+#  1: number of minimum worker nodes
+function scale_worker_nodes() {
+  echo "Reconcile workers to at least ${1} nodes"
```
mgencur, line 5: Maybe it would be good to put this line here. It would force the user to pass the argument
```
num_workers=${1:?Pass num_workers as arg[1]}
```

Aside that, was there any agreement on having all stuff in this repo in Python? If not, I'd just merge this and use it as is.

mgencur, line 5: Maybe it would be good to put this line here. It would force the user to pass the argument
```
num_workers=${1:?Pass num_workers as arg[1]}
```

Aside that, was there any agreement on having all stuff in this repo in Python? If not, I'd just merge this and use it as is.

pierDipi: Done.

I don't think there was an agreement on using python, in this case, we would be required to use Kube clients and manage dependencies, etc, not ideal for this case.

pierDipi: GH created 2 comments with the same text? https://github.com/openshift-knative/hack/pull/1#discussion_r890037617

pierDipi:
Merging manually.

PR #2, opened 2022-07-11
Openshift/release reconciler

pierDipi:
TODO:
- [ :heavy_check_mark: ] Image mirroring


generate_ci.py
```diff
@@ -0,0 +1,392 @@
...
+        }
+    },
+}
+
+# TODO take them from SO main?
+#  There might be a delay between having CI cluster pools for a max version and our support version for OCP, so let's
+#  not do this until we are sure about getting cluster pools and supported versions.
+openshift_versions = [
+    "4.6",  # min version
+    "4.10"  # max version
```
matzew, line 100: no 4.11 pool yet :sweat_smile:

dsimansk:
I might be missing something obvious (as a python noob). Is there any benefit writing it in `.py` instead of `.go`? `go install/go run` seems like a nice tool to distribute and reuse such tool.

generate_ci.py
```diff
@@ -0,0 +1,416 @@
...
+# 2. Dockerfiles are in the directory matching
+#    `openshift/ci-operator/knative-images/**/Dockerfile`
+#    or openshift/ci-operator/knative-test-images/**/Dockerfile
+# 3. The repository has a Makefile in the root and has some targets containing the keyword
+#    `test`
+
+# Usage:
+# ./generate_ci.py \
+#   --openshift-release-remove git@github.com:pierDipi/release.git # Remote to your fork
+
```
aliok, line 24: @pierDipi I've read this but I still don't understand what this script does.

Can you write some explanation please? Then I can review

pierDipi: it generates openshift/release configuration for configured repositories (repo variable) and supported branches.

pierDipi: when we create a new branch in a midstream repository, the usual flow looks like this:

1. mirror upstream branch
2. apply openshift directory from main
3. go to openshift/release and add CI jobs and image mirroring for that branch

I'm automating 3 for _every_ midstream repository (except SO for reasons), and we will eventually try to automate 1 and 2 in this automation

generate_ci.py
```diff
@@ -0,0 +1,416 @@
...
+releases:
+  initial:
+    integration:
+      name: "{ov}"
+      namespace: ocp
+  latest:
+    integration:
+      include_built_images: true
+      name: "{ov}"
+      namespace: ocp
```
dsimansk, line 350: AFAIK this whole part doesn't have any effect since we started to rely on `cluster_claim` pools. Those should control and define OCP version etc.

pierDipi: Thanks!

pierDipi: which part exactly?

pierDipi: the entire?
```
releases:
  initial:
    integration:
      name: "{ov}"
      namespace: ocp
  latest:
    integration:
      include_built_images: true
      name: "{ov}"
      namespace: ocp
```

generate_ci.py
cardil: You could reuse a tool I wrote couple of months ago https://github.com/cardil/deviate

It's used for kn-plugin-event. Here's usage https://github.com/openshift-knative/kn-plugin-event/blob/main/openshift/release/update-to-head.sh#L9-L11

pierDipi: @cardil for 1 and 2 right?

pierDipi: we discussed your tool in our weekly call a while ago and it was one of the options we considered for 1 and 2.

pierDipi: I like it, can you please showcase how we can use it in `openshift-knative/eventing-kafka-broker` or `openshift/knative-eventing`?

pierDipi: because I see that the sync'ed branches are a bit different from what we have currently like:
upstream `release-1.5` -> midstream `release-v1.5`.

I'm sure we can add a feature to add the `v` if it's not there already.

skonto:
Not a strict requirement but some improvements can be done to make code more Python-like as [pylint complains](https://gist.github.com/skonto/f60c90ac855a2e81f09325455fd941c5) for several stuff.

generate_ci.py
cardil: Those branch names and tags are configurable. It should be possible to use it after configuration.

I plan the Deviate tool could generate appropriate configs for openshift/release...

generate_ci.py
```diff
@@ -0,0 +1,416 @@
...
+            file_name = f"mapping_{promotion_name}_{repo_name}_quay"
+            with open(f"{openshift_release_image_mirroring}/{file_name}", "w") as f:
+                f.writelines(mirroring_lines)
+
+            images_field = ""
+            dependencies = ""
+            for img in images + test_images:
+                images_field += f"""
+- dockerfile_path: openshift/{img.path}/Dockerfile
+  to: {img.name}"""
```
dsimansk, line 287: I've notice this "auto-fix" PRs.  https://github.com/openshift/release/commit/eb4d36ee1d223e13d75dfbd2d823778a2aedc708
IMO it's driven from whatever is detected in Dockerfile. https://github.com/openshift/knative-eventing/blob/release-next/openshift/ci-operator/knative-images/apiserver_receive_adapter/Dockerfile

The most simple solution would be to either define add similar builder stream to `base_images` and then`from:` here:
```
- dockerfile_path: openshift/ci-operator/knative-images/kafka-controller/Dockerfile
  from: <builder_stream_name>
  to: knative-eventing-kafka-broker-kafka-controller
```

It can be optimized a bit to to pre-build those binaries at once and copy them over to images. And stop running `make install test-install` for every image build.

generate_ci.py
dsimansk: Yes, whole block is L341-350.

generate_ci.py
```diff
@@ -103,22 +110,20 @@
...
+    },
+    "release-v1.5": {
+        ocp_min_version_key: "4.8",
+        ocp_max_version_key: "4.11",
+    },
+    "release-next": {
+        ocp_min_version_key: "4.8",
+        ocp_max_version_key: "4.11",
+    }
+}
```
matzew, line 126: this is neat structure

matzew:
I think this looks good now...

matzew (approved):
/lgtm
/approve

/hold
for others - but feel free to unhold.

This has had several iterations, and latest PR generated for required OCP updates looks just good

PR #3, opened 2022-10-04
Github action for multiarch test images

mgencur:
Part of https://issues.redhat.com/browse/SRVCOM-2053

This is a reusable github action which builds all test images from the given repository for multiple architectures and pushes them to a predefined repository structure in quay.io

The quay username and password will be defined at organization level and passed to the action.

This is used in https://github.com/openshift-knative/eventing-kafka-broker/pull/420

### DIFF
--- a/.github/workflows/multiarch-build.yaml
+++ b/.github/workflows/multiarch-build.yaml
@@ -1,0 +1,48 @@
+name: Multiarch builds
+
+on:
+  workflow_call: {}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  multiarch-build:
+    name: Build test images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
+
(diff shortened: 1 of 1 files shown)


mgencur:
/hold
I want to investigate how to pass the correct Golang version. We don't want to be changing this reusable action with every release.

mgencur:
/hold cancel
Maybe it's good enough to have the go version defined here. The upstream reusable knative action defines it as well: https://github.com/knative/actions/blob/main/.github/workflows/go-build.yaml

.github/workflows/multiarch-build.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  multiarch-build:
+    name: Build test images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
```
pierDipi, line 18: We can probably already go with go 1.18

pierDipi (approved):
/lgtm
/approve

PR #4, opened 2022-10-11
Improve parsing major and minor version from branch name in github ac…

mgencur:
…tion

* It can now work with all styles: release-vX.Y.Z release-X.Y.Z release-X.Y

This is to make it work with https://github.com/openshift-knative/kn-plugin-event which has different naming style for branches.

### DIFF
--- a/.github/workflows/multiarch-build.yaml
+++ b/.github/workflows/multiarch-build.yaml
@@ -42,7 +42,7 @@ jobs:
           ref: ${{ github.ref_name }}
         run: |
           echo "Building images from ${GITHUB_REF}@${GITHUB_SHA}"
-          # Remove release-v prefix and parse major and minor version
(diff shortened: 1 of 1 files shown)


.github/workflows/multiarch-build.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           ref: ${{ github.ref_name }}
         run: |
           echo "Building images from ${GITHUB_REF}@${GITHUB_SHA}"
-          # Remove release-v prefix and parse major and minor version
-          major_minor=$(echo ${ref#release-v} | awk -F \. '{printf "%d.%d", $1, $2}')
+          # Ref: https://regex101.com/r/Po1HA3/1
```
pierDipi, line 45: Will this link go away?
Can we add that examples as part of the comment?

pierDipi (approved):
/lgtm

PR #6, opened 2022-11-04
Rewrite reconciler in Go

pierDipi:
This was originally asked and I agree.

The nice properties of having this tool in Go is that we can
reuse openshift/ci-tools types, leading to a much cleaner
and more reliable tool.

In particular, I wanted to onboard eventing-hyperfoil-benchmark
but the existing reconciler failed because of some naming
limitations in openshift/release, this new reconciler handles
those cases by using a predictable way of shortening test
names.

To run it, you can execute:
```
make generate-ci REMOTE=git@github.com:pierDipi/release.git
```

The configurations are in the `config/` directory.

Also, with this little rewrite we can add more automation
and it is much easier because we don't need to grep
and replace via regex on yaml files (think about the
`always_run` problem we have discussed).


config/eventing.yaml
```diff
@@ -0,0 +1,40 @@
...
+  branches:
+    "release-v1.4":
+      openShiftVersions:
+        - 4.11
+        - 4.6
+    "release-v1.5":
+      openShiftVersions:
+        - 4.11
+        - 4.8
+    "release-v1.6":
```
matzew, line 13: can we add the `release-v1.7` branch here too ?

matzew:
I've added `CONTAINER_ENGINE` to my env, and upate the go file to be just like `make update`, I am now getting lot's of the below:

```
...
time="2022-12-19T12:58:15Z" level=error error="output tag openshift/knative-eventing-mtchannel-broker:knative-v1.4 is promoted from more than one place: openshift/knative-eventing@release-v1.4 [411], openshift-knative/eventing@release-v1.4 [411]"
...
```

Promotion should be disabled on all configs, expect for one (per version)

matzew:
I was able to get this: 

https://github.com/openshift/release/pull/34919/files

matzew (approved):
/lgtm
/approve

PR #8, opened 2022-11-30
:gift: Containerfile multi-arch build

cardil:
## Changes

 - :gift: A multi-arch Containerfile image build using buildah

### DIFF
--- a/.github/workflows/multiarch-build.yaml
+++ b/.github/workflows/multiarch-build.yaml
@@ -1,4 +1,4 @@
-name: Multiarch builds
+name: Multiarch KO builds
 
 on:
   workflow_call: {}
@@ -9,7 +9,7 @@ defaults:
 
 jobs:
   multiarch-build:
-    name: Build test images
+    name: Build images
     runs-on: ubuntu-latest
     steps:
       - name: Setup Golang
--- a/.github/workflows/multiarch-containerfile-build.yaml
+++ b/.github/workflows/multiarch-containerfile-build.yaml
@@ -1,0 +1,69 @@
+name: Multiarch Buildah builds
+
+on:
+  workflow_call:
+    inputs:
+      image:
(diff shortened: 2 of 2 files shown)


mgencur:
I think this github action looks ok in general. 
The usage example in https://github.com/cardil/kn-plugin-event/blob/multiarch/release-1.5/.github/workflows/multiarch-build.yaml will probably need some tweaks:
* push to a different repo, under https://quay.io/repository/openshift-knative/kn-plugin-event/, e.g. https://quay.io/repository/openshift-knative/kn-plugin-event/wathola-forwarder
* it duplicates some dockerfiles which are already in that repo under openshift/ dir

cardil:
> push to a different repo, under https://quay.io/repository/openshift-knative/kn-plugin-event/, e.g. https://quay.io/repository/openshift-knative/kn-plugin-event/wathola-forwarder

There is a `registry` input. Production value would propably be simply: `quay.io`, which result in images like: `quay.io/openshift-knative/kn-plugin-event/test/wathola-forwarder`

---

> it duplicates some dockerfiles which are already in that repo under openshift/ dir

I looked at them, and I think the `openshift/` container files couldn't be reused because they are using Openshift specific things like:

```Dockerfile
FROM base
```

---

> doublecheck if the binaries are really built for the specific platforms. In my opinion, the binaries are always built for amd64 and only the image metadata are specific for the different platforms.

Yup, we need to double-check that. I don't know how, though. But, I think it is a real build. See this previous build, where I did try to build images for `linux/s390x` as well. It failed, for that platform: https://github.com/cardil/kn-plugin-event/actions/runs/3583459982/jobs/6028914644

cardil:
There is no [`OWNERS`](https://github.com/openshift-knative/hack/blob/main/OWNERS) files. We need to merge manually.

PR #12, opened 2022-12-07
:gift_heart: Auto-skip push to registry if secrets are not set

cardil:
## Changes

 - :gift_heart: Auto-skip push to registry, if secrets are not set
   - Allows to build images, checking the build

/kind enhancement

### DIFF
--- a/.github/workflows/multiarch-containerfile-build.yaml
+++ b/.github/workflows/multiarch-containerfile-build.yaml
@@ -60,6 +60,7 @@ jobs:
 
       - name: Push To ${{ inputs.registry }}
         id: push-to-quay
+        if: secrets.QUAY_USER && secrets.QUAY_PASSWORD
         uses: redhat-actions/push-to-registry@v2
         with:
           image: ${{ steps.build-image.outputs.image }}
--- a/.github/workflows/multiarch-containerfile-build.yaml
+++ b/.github/workflows/multiarch-containerfile-build.yaml
@@ -30,6 +30,8 @@ jobs:
   multiarch-build:
     name: Build images
     runs-on: ubuntu-22.04
+    env:
+      PRINCIPAL: '${{ secrets.QUAY_USER }}:${{ secrets.QUAY_PASSWORD }}'
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


cardil:
Builds well: https://github.com/cardil/kn-plugin-event/actions/runs/3639317955/jobs/6142552634

/cc @mgencur 
/cc @pierDipi

pierDipi (commented):
I lack context, what's the reason for this?

pierDipi:
> I would like to have option to run those builds for PRs, and other development branches, so you may know that it should build the images well, before the merge.

In that case, can't we use something like `if: github.event_name == 'push'`?
it seems like an hack to use secrets presence to infer push or not push

mgencur:
What does it bring ... depending on the secrets? This github action will always run on github, not locally.
It's more understandable to have what @pierDipi suggested. The user knows that pushing will be done only on push, not in pre-submit phase. If you want this for debugging purposes to run this on your repo then pushing the images will fail anyway if you don't have the secrets set, which might be a good signal. In other words, failing the github action if the secrets are not provided is expected.
Depending on the presence of secrets only raises questions like the one from @pierDipi . Using `if: github.event_name == 'push'` makes it clear immediately.

cardil:
@mgencur wrote:

> What does it bring ... depending on the secrets? This github action will always run on github, not locally.

Sure, it will always run on Github. I thought this may allow testing the build on fork, as the forks doesn't have secrets configured. You push your code to your fork, and Github action fires on your fork repo, before you even create a PR.

> It's more understandable to have what @pierDipi suggested. The user knows that pushing will be done only on push, not in pre-submit phase. If you want this for debugging purposes to run this on your repo then pushing the images will fail anyway if you don't have the secrets set, which might be a good signal. In other words, failing the github action if the secrets are not provided is expected. Depending on the presence of secrets only raises questions like the one from @pierDipi . Using `if: github.event_name == 'push'` makes it clear immediately, it is self-documenting.

I think both of those conditions are valid. I'll add `github.event_name == 'push'`. That would allow to run this workflow for the [`pull_request` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request).

dsimansk:
Superseded by Konflux based pipelines and workflows. 

/close

PR #13, opened 2022-12-20
Add knative eventing 1.8 branch

creydr:
Added knative eventing 1.8 branch.

In addition, I separated the two config of the kafka-broker and eventing as sometimes (as now) the branches between those are not aligned.


Makefile
```diff
@@ -1,6 +1,7 @@
 generate-ci:
```
matzew, line 1: should we perhaps create dedicated make targets ? 

`eventing-core`, `ekb`, or the like?

matzew (approved):
/lgtm
/approve

PR #14, opened 2022-12-20
Adding transformer function for knative-eventing-src image

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

Currently all `knative-eventing-src` references are dead. The new tool just uses the name of the repo (`eventing`), but we MUST maintain the `knative-eventing`, to not break older Serverless Operator branches.

Proposal: add a temporary "transformer". 

In the long run the tool should have options:
* do just eventing
* do just eventing-kafka-broker
* do just a certain version of a repository
etc


creydr (approved):
/lgtm

PR #16, opened 2023-01-03
Moving firstVersion to be disabled promotion, instead of 'by name'

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

The code diff was already used to generate the content of: https://github.com/openshift/release/pull/35106

### DIFF
--- a/pkg/prowgen/prowgen_config.go
+++ b/pkg/prowgen/prowgen_config.go
@@ -115,9 +115,9 @@ func NewGenerateConfigs(ctx context.Context, r Repository, cc CommonConfig, opts
 			copy(options, opts)
 			if isFirstVersion {
 				isFirstVersion = false
-				options = append(options, withNamePromotion(r, branchName))
+				options = append(options, withTagPromotion(r, branchName, true))
 			} else {
-				options = append(options, withTagPromotion(r, branchName))
+				options = append(options, withTagPromotion(r, branchName, false))
 			}
 
 			options = append(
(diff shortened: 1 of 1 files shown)


pkg/prowgen/prowgen_config.go
```diff
@@ -115,9 +115,9 @@ func NewGenerateConfigs(ctx context.Context, r Repository, cc CommonConfig, opts
 			copy(options, opts)
 			if isFirstVersion {
 				isFirstVersion = false
-				options = append(options, withNamePromotion(r, branchName))
+				options = append(options, withTagPromotion(r, branchName, true))
 			} else {
-				options = append(options, withTagPromotion(r, branchName))
+				options = append(options, withTagPromotion(r, branchName, false))
```
pierDipi, line 120: We could just remove the `else { ... }` ?

matzew: and skip any other generation? because we remove the entire `else` block?

pierDipi: no, no skip, it will keep going but only without the `promotion` filed when `!firstVersion`

pierDipi: I mean only this else block
```golang
else {
				options = append(options, withTagPromotion(r, branchName))
}
```

matzew:
/close

for now we close it, as this could (likely will) lead to unintended consequences of removing images w/ the "old" named patterns, ... which we still reference, somewhere (eg. RHOSE)

PR #20, opened 2023-02-02
Fix mirroring regex

pierDipi:
The regex for eventing was deleting the ekb mirroring files.


creydr (approved):
/lgtm

https://github.com/openshift/release/compare/master...creydr:openshift-release:sync-serverless-ci-cp?expand=1

(files are not empty - only permissions changed)

PR #28, opened 2023-03-09
Set OCP min version to 4.10 for knative eventing >=v1.8

creydr:
As per title


creydr:
/assign @matzew @pierDipi 
Here the one to update the prowgen configs

pierDipi (approved):
/lgtm
/approve

PR #31, opened 2023-03-13
Add dockerfile generator

pierDipi:
This new tool will help generating dockerfiles for
Go based components + generate an images.yaml
file reference for rekt based tests (and more ...).

See `pkg/project/testdata` and `pkg/project/testoutput`
for an example of generating dockerfiles + image
reference for hack repository itself.


pierDipi:
for eventing-istio a need a file that references eventing and eventing kafka broker test images similar to https://github.com/openshift-knative/eventing/blob/release-next/openshift/images.yaml, so generating everything is helpful to maintain those mappings.

In eventing istio, I will download those files from the matching branches

cmd/generate/Dockerfile.template
```diff
@@ -0,0 +1,16 @@
+# DO NOT EDIT! Generated Dockerfile for {{.main}}.
+FROM {{.builder}} as builder
+
+COPY . .
+
+RUN mkdir -p /var/run/ko && \
```
matzew, line 6: instead of the `make` targets, we will run this (`go build`) directly, rather than via the `Makefile`?

Is that the intention?

cmd/generate/generate.go
```diff
@@ -0,0 +1,285 @@
...
+	GenerateDockerfileOption = "dockerfile"
+)
+
+//go:embed Dockerfile.template
+var DockerfileTemplate embed.FS
+
+//go:embed BuildImageDockerfile.template
+var DockerfileBuildImageTemplate embed.FS
+
+func main() {
```
matzew, line 38: very long, but not really a concern. Let's get this out, and improve later

cmd/generate/generate.go
```diff
@@ -0,0 +1,285 @@
...
+					goPackageToImageMapping[filepath.Join(goMod.Module.Mod.Path, p)] = image
+				}
+			}
+		}
+
+		mapping, err := yaml.Marshal(goPackageToImageMapping)
+		if err != nil {
+			log.Fatal(err)
+		}
+		if err := os.WriteFile(filepath.Join(output, "images.yaml"), mapping, fs.ModePerm); err != nil {
```
matzew, line 253: what is images.yaml?  Can we have some comments on why/what?

pkg/project/testdata/project.yaml
```diff
@@ -0,0 +1,3 @@
+project:
```
matzew, line 1: is this inspired by the `project.yaml` from SO? 

What's the file for? Let's document it, somehow

pkg/project/testoutput/images.yaml
```diff
@@ -0,0 +1,3 @@
+github.com/openshift-knative/hack/cmd/generate: hello
```
matzew, line 1: Not sure I fully understand this file, why/what

matzew:
this tool/addion will update the midstream operand repos, right?

While originally this tool updates  CI repo (openshift/release)

If so: perhaps two binaries?  or some better separation?

pierDipi:
> this tool/addion will update the midstream operand repos, right?
> 

Yes, it will eventually help onboarding SO to prowgen as well

> While originally this tool updates CI repo (openshift/release)
> 
> If so: perhaps two binaries? or some better separation?

They are 2 separate binaries (that work well with each other)

pkg/project/testoutput/images.yaml
pierDipi: Added comments on the code that generates it.

pkg/project/testdata/project.yaml
pierDipi: Added a few comments

cmd/generate/generate.go
pierDipi: Yes, not the cleanest code

cmd/generate/Dockerfile.template
pierDipi: Yes, this will speed up the process of building the images, since not every image will rebuild every binary as currently done with `make install`

PR #41, opened 2023-03-20
Real unique names for tests now

### LINKED ISSUE #40: SHA1 sum for unique name generation doesn't really generate unique stuff
See https://go.dev/play/p/bcGg01t0fu2

We do 
```
hex.EncodeToString(sha1.New().Sum([]byte("foo)))
```

but instead, we should be doing:
```
h := sha1.New()
h.Write([]byte("foo))
hex.EncodeToString(h.Sum(nil))
```

Using the first method
```
sha1 of test-kafka-broker-upstream-nightly                                                                   is 746573742d6b61666b612d62726f6b65722d757073747265616d2d6e696768746c79da39a3ee5e6b4b0d3255bfef95601890afd80709
sha1 of test-kafka-broker-namespaced-midstream-nightly-ondemand                                              is 746573742d6b61666b612d62726f6b65722d6e616d657370616365642d6d696473747265616d2d6e696768746c792d6f6e64656d616e64da39a3ee5e6b4b0d3255bfef95601890afd80709
```

Using the 2nd method
```
sha1 of test-kafka-broker-upstream-nightly                                                                   is fbbddbf654e28a849850c4d3bb208941d3d170ca                                                            
sha1 of test-kafka-broker-namespaced-midstream-nightly-ondemand                                              is 829372326521d7395acf9b08228dbf80c4b8ae2c 
```

Well, the first one also generates something unique, but we only use the first 7 chars and they don't become unique.

aliok:
Fixes #40

### DIFF
--- a/pkg/prowgen/prowgen_command_name_mapping_test.go
+++ b/pkg/prowgen/prowgen_command_name_mapping_test.go
@@ -26,7 +26,7 @@ func TestToName(t *testing.T) {
 				Command: strings.Repeat("a", maxNameLength),
 			},
 			openShiftVersion: openshiftVersion,
-			want:             fmt.Sprintf("%s-%s%s", strings.Repeat("a", maxNameLength-len(suffix)-len(continuousSuffix)-shaLength-1) /* hex sha1 */, "6161616", suffix),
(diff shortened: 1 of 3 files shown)


pierDipi (approved):
/lgtm

PR #43, opened 2023-03-30
use openshift-knative/hack for serving

ReToCode:
This PR enables Serving to use `openshift-knative/hack`. Jobs were generated here: https://github.com/openshift/release/pull/35397. Before we merge https://github.com/openshift/release/pull/35397, I'd like some feedback on these changes.

## Changes
- Introduce a new field in `config` to override the default `cron` expression (Serving [used](https://github.com/openshift-knative/serving/blob/main/openshift/ci-operator/generate-ci-config.sh#L47) a different cron per branch)
- Introduce a new field `imageNameOverrides` to override generated docker image names (Eventing [seems](https://github.com/openshift-knative/eventing/blob/main/Makefile#L21) to build storage-version-migration twice to have a matching golang binary. This can be dropped with this change).
- Changes the generation of the image mirror files to [avoid](https://redhat-internal.slack.com/archives/CD87JDUB0/p1680071453233949?thread_ts=1680070041.097729&cid=CD87JDUB0). The generated mirror-file look like [this](https://github.com/openshift/release/pull/35397/files#diff-5b26525b9d9d467a6302c33ae6e0ec661f3ab71d666148ea4218eb45c39f9441R1) instead of [this](https://github.com/openshift/release/blob/master/core-services/image-mirroring/knative/mapping_knative_knative-v1.8_eventing_quay#L1) (the source has a specific tag). 
- Adds Serving config with the same override for legacy repo naming as Eventing uses.

/cc @nak3, @pierDipi, @skonto, @matzew

### DIFF
--- a/config/serving.yaml
+++ b/config/serving.yaml
@@ -1,0 +1,41 @@
+# Full struct in cmd/prowgen/prowgen.go#Config
+
+config:
+  branches:
+    "release-v1.7":
+      cron: "0 17 * * 1-5"
+      openShiftVersions:
+        - 4.12
+        - 4.8
+    "release-v1.8":
+      cron: "0 19 * * 1-5"
+      openShiftVersions:
+        - 4.12
+        - 4.10
+    "release-v1.9":
+      cron: "0 21 * * 1-5"
+      openShiftVersions:
+        - 4.12
+        - 4.10
+    "release-next":
+      cron: "0 8 * * 1-5"
+      openShiftVersions:
+        - 4.12
+        - 4.10
+
+repositories:
+  - org: openshift-knative
+    repo: serving
+    imagePrefix: knative-serving
+    imageNameOverrides:
+      migrate: storage-version-migration
+    slackChannel: "#knative-serving-ci"
+    e2e:
+      matches:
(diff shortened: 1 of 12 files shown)


ReToCode:
I'll generate a second PR to see what would change for Eventing with those changes.

nak3:
LGTM :tada:

config/serving.yaml
```diff
@@ -0,0 +1,41 @@
...
+    slackChannel: "#knative-serving-ci"
+    e2e:
+      matches:
+        - ".*e2e$"
+        - ".*e2e-tls$"
+    resources:
+      '*':
+        requests:
+          cpu: 500m
+          memory: 4Gi
```
pierDipi, line 41: Is this actually needed? Like 4Gi? :exploding_head:

ReToCode: I haven't checked, but all our builds are currently running with that. I would prefer to keep the current value to not introduce a potential flake while trying to get our builds running again with the new image tag pattern.

pierDipi (approved):
/lgtm
/approve

PR #49, opened 2023-05-08
Install missing zoneinfo into to ubi-minimal image

matzew:
The `zoneinfo` is missing on UBI Minimal images, but the `PingSource` makes usage of that, hence installing it to our midstream images


pierDipi (approved):
/lgtm
/approve

PR #56, opened 2023-05-30
Use openshift-knative/hack for net-* repositories

ReToCode:
- Use openshift-knative/hack for net-* repositories
- Do not fail if no `Makefile` is defined in the repository

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ generate-ci:
 generate-serving-ci:
 	rm -rf openshift openshift-knative
 	go run github.com/openshift-knative/hack/cmd/prowgen --config config/serving.yaml --remote $(REMOTE)
+	go run github.com/openshift-knative/hack/cmd/prowgen --config config/serving-net-istio.yaml --remote $(REMOTE)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -14,6 +14,8 @@ generate-ci:
 generate-serving-ci:
 	rm -rf openshift openshift-knative
 	go run github.com/openshift-knative/hack/cmd/prowgen --config config/serving.yaml --remote $(REMOTE)
+	go run github.com/openshift-knative/hack/cmd/prowgen --config config/serving-net-istio.yaml --remote $(REMOTE)
+	go run github.com/openshift-knative/hack/cmd/prowgen --config config/serving-net-kourier.yaml --remote $(REMOTE)
```
pierDipi, line 18: Can we add these 2 to `generate-ci`? or alternatively we could define `generate-ci` to run `generate-eventing-ci` and `generate-serving-ci` to avoid the duplication

ReToCode: Sure, I added it as duplication as the `rm -rf openshift openshift-knative` part makes it harder to depend on another target. Alternative is to have more conditionals somehow to make both cases work.

pierDipi (approved):
/lgtm
/approve

PR #57, opened 2023-06-26
Introduce Prowcopy to generate Prow config starting from another branch config

pierDipi:
`prowcopy` allows to create Prow configuration in openshift/release starting from the configuration
of another branch. For example, if you have configuration for CI of the main branch, and then you
cut a new branch, the main branch configuration can be used to define CI of the new branch.

For now, it is specifically tailored for serverless-operator.

This is to be able to automate PRs like: https://github.com/openshift/release/pull/40491 using
workflows like: https://github.com/pierDipi/serverless-operator/blob/main/.github/workflows/release-generate-ci.yaml

See usage in https://github.com/openshift-knative/serverless-operator/pull/2131

PR #58, opened 2023-06-27
Use OCP 4.13 for serving midstream CI

nak3:
This patch changes to use OCP 4.13 for serving midstream CI.
The configfiles generated by this change is adding via https://github.com/openshift/release/pull/40674

/cc @skonto @ReToCode


nak3:
Unit test is broken now regardless of this change - https://github.com/openshift-knative/hack/pull/57#issuecomment-1608850097

PR #65, opened 2023-07-04
Reverting the CI for release-v1.10 to openshift 4.12

Cali0707:
It looks like the CI is failing on openshift 4.13 (potentially because of cert-manager GA). We will open a ticket to investigate this, but for now should revert to 4.12 so that there can be tests running on the `release-v1.10` branch.


pierDipi (approved):
/lgtm
/approve

PR #68, opened 2023-08-01
Add eventing 1.11 family

creydr:
Adds eventing core, EKB and eventing-istio v1.11 branch for OCP 4.10 & (~~4.12~~) 4.13

### DIFF
--- a/config/eventing.yaml
+++ b/config/eventing.yaml
@@ -22,6 +22,10 @@ config:
       openShiftVersions:
         - 4.12
         - 4.10
+    "release-v1.11":
+      openShiftVersions:
+        - 4.13
+        - 4.10
     "release-next":
       openShiftVersions:
         - 4.12
--- a/config/eventing-istio.yaml
+++ b/config/eventing-istio.yaml
@@ -14,6 +14,10 @@ config:
       openShiftVersions:
         - 4.13
         - 4.10
+    "release-v1.11":
+      openShiftVersions:
+        - 4.13
+        - 4.10
     "release-next":
       openShiftVersions:
         - 4.12
--- a/config/eventing-kafka-broker.yaml
+++ b/config/eventing-kafka-broker.yaml
@@ -22,6 +22,10 @@ config:
       openShiftVersions:
         - 4.12
         - 4.10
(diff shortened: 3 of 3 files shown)


pierDipi (commented):
Can we add others as well?

config/eventing.yaml
```diff
@@ -22,6 +22,10 @@ config:
       openShiftVersions:
         - 4.12
         - 4.10
+    "release-v1.11":
+      openShiftVersions:
+        - 4.13
```
pierDipi, line 27: 4.13 won't work due to cert manager subscription being TP

creydr:
> Can we add others as well?

I just wanted to keep it separated.

config/eventing.yaml
creydr: Could you clarify a bit on it?

pierDipi: There isn't a single subscription for cert-manager that works on 4.10 as well as on 4.13

pierDipi: that's also the reason why we are still using 4.12 on release-next

creydr: The eventing-istio tests are running on 4.13: https://github.com/openshift-knative/hack/blob/202e76d406d5d0dfb45e066bd09c3b97c8b240b6/config/eventing-istio.yaml#L13-L16 and use cert-manager (via SO) too: https://github.com/openshift-knative/eventing-istio/blob/ca3dc59e4d13c848c73c7f70a49f8546a26384ef/openshift/install.sh#L12

Eventing seems to use the same from SO: https://github.com/openshift-knative/eventing/blob/f1b58f234e8a262cee577e5f806428c3b532141e/openshift/e2e-common.sh#L91C114-L91C130 and https://github.com/openshift-knative/serverless-operator/blob/f97208bc569f1d45d135f5887ad444d6bc69ccfe/Makefile#L31-L33

What am I missing?

creydr:
> > Can we add others as well?
> 
> I just wanted to keep it separated.

@pierDipi updated to include EKB & istio too

pierDipi (approved):
/lgtm

PR #69, opened 2023-08-03
Drop cron config from serving

nak3:
As per title, cron config is modified by release repo's bot.
So this patch drops it.

Note, eventing and other config do not have it.


pierDipi (approved):
/lgtm
/approve

PR #70, opened 2023-08-04
Bump OCP min version for release-next to 4.11

nak3:
OCP 4.10 is EOL on September 10, 2023 as per https://access.redhat.com/support/policy/updates/openshift#dates so release-next (serving 1.12) should not need to test on 4.10.

(Also, net-kourier introduced the grpc probe so we cannot use 4.10).

/cc @ReToCode @skonto


ReToCode (approved):
/lgtm
/approve

PR #76, opened 2023-08-31
Update ocp for recent knative eventings

matzew:
As per title

With 1.11 we do support 4.11+

Hence updating:
* 1.11
* release-next (and aligning to what serving does here)


pierDipi (approved):
/lgtm
/approve

PR #79, opened 2023-10-17
[WIP] SRVCOM-2762: Add step for snyk security scanning

creydr:
Adding a step for the snyk security scanning (https://issues.redhat.com/browse/SRVCOM-2762)

### DIFF
--- a/pkg/prowgen/prowgen_tests.go
+++ b/pkg/prowgen/prowgen_tests.go
@@ -94,6 +94,7 @@ func DiscoverTests(r Repository, openShiftVersion string, cronOverride *string) 
 					Workflow: pointer.String("generic-claim"),
 				},
 			}
+
 			cfg.Tests = append(cfg.Tests, testConfiguration)
 
 			cronTestConfiguration := testConfiguration.DeepCopy()
@@ -106,6 +107,20 @@ func DiscoverTests(r Repository, openShiftVersion string, cronOverride *string) 
 
 			cfg.Tests = append(cfg.Tests, *cronTestConfiguration)
 		}
+
+		securityTestConfiguration := cioperatorapi.TestStepConfiguration{
+			As:       "security",
+			Optional: true,
+			MultiStageTestConfiguration: &cioperatorapi.MultiStageTestConfiguration{
+				Environment: cioperatorapi.TestEnvironment{
+					"PROJECT_NAME": r.Repo,
+					"ORG_NAME":     r.Org,
+				},
+				Workflow: pointer.String("openshift-ci-security"),
+			},
+		}
+
+		cfg.Tests = append(cfg.Tests, securityTestConfiguration)
 
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


pkg/prowgen/prowgen_tests.go
```diff
@@ -107,6 +108,20 @@ func DiscoverTests(r Repository, openShiftVersion string, cronOverride *string)
...
+			MultiStageTestConfiguration: &cioperatorapi.MultiStageTestConfiguration{
+				Environment: cioperatorapi.TestEnvironment{
+					"PROJECT_NAME": r.Repo,
+					"ORG_NAME":     r.Org,
+				},
+				Workflow: pointer.String("openshift-ci-security"),
+			},
+		}
+
+		cfg.Tests = append(cfg.Tests, securityTestConfiguration)
```
pierDipi, line 123: This will add one test per openshift version and per branch, I was wondering if we can that option one level above

in Prowgen config something like this:
```golang
			if isFirstVersion {
				isFirstVersion = false
				options = append(options, withNamePromotion(r, branchName))

				options = append(options, WithSnykScanning(r, branchName))
			} else {
				options = append(options, withTagPromotion(r, branchName))
			}
```

Also, how does it distinguish between branches?

pkg/prowgen/prowgen_tests.go
```diff
@@ -107,6 +108,20 @@ func DiscoverTests(r Repository, openShiftVersion string, cronOverride *string)
 			cfg.Tests = append(cfg.Tests, *cronTestConfiguration)
 		}
 
+		securityTestConfiguration := cioperatorapi.TestStepConfiguration{
+			As:       "security",
+			Optional: true,
+			MultiStageTestConfiguration: &cioperatorapi.MultiStageTestConfiguration{
+				Environment: cioperatorapi.TestEnvironment{
+					"PROJECT_NAME": r.Repo,
+					"ORG_NAME":     r.Org,
```
pierDipi, line 117: Do we need to create the org before? How does it get access to it?

pierDipi (commented):
Thanks @creydr

pkg/prowgen/prowgen_tests.go
pierDipi: Altenatively we could create a more complex project name which includes the repository org `<org>-<repo>` and maybe also branch suffix so that we have the analysis in the existing openshift org per branch instead of managing our own org, but I'm not sure what's best, I'd connect with TP team to see what they suggest and the undocumented mechanics

pierDipi: I see the test run PR failure so I suspect there is something to do in this area

creydr: Yes. I just saw that `If your organization is not OpenShift, you can optionally override the ORG_NAME environment variable to point to your organization in snyk UI.`
Have to check how to create that org and how it gets accessed

pierDipi: Let's ask them in #forum-ocp-testplatform

pierDipi: You already did that! :fastparrot:

pierDipi:
We have snyk running in Konflux now

PR #80, opened 2023-10-25
Make generate-groups.sh and generate-knative.sh executable

mgencur:
Place this in "src" template which is guaranteed to have the right version of Eventing checked out and workdir properly set. Upstream scripts make the files executable but they would fail because CI pods do not run with root access and don't have permissions to change ownership.

### DIFF
--- a/cmd/generate/SourceImageDockerfile.template
+++ b/cmd/generate/SourceImageDockerfile.template
@@ -1,3 +1,5 @@
 # DO NOT EDIT! Generated Dockerfile.
 
 FROM src
+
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm
/approve

pierDipi:
This needs an update as it's used in the test https://github.com/openshift-knative/hack/blob/main/pkg/project/testoutput/openshift/ci-operator/source-image/Dockerfile

mgencur:
Wait. I need to consider the case when the files do not exist because this is a common src image for all Serverless images.

pierDipi (approved):
/lgtm
/approve

PR #81, opened 2023-10-26
Enable custom source image builds

mgencur:
* The custom "src" image is built similar to any other image, allows for parsing image inputs and base images.
* The only difference for "src" image is that it needs to define "from: src" in CI Operator config instead of using the src image in inputs. Otherwise CI Operator fails to properly setup context paths.

We already use a similar approach for serverless-operator repo where we have "serverless-operator-src" image. So, this PR is also a preparation for onboarding serverless-operator repo to automatic CI generation.

### DIFF
--- a/pkg/prowgen/prowgen_config.go
+++ b/pkg/prowgen/prowgen_config.go
@@ -128,11 +128,11 @@ func NewGenerateConfigs(ctx context.Context, r Repository, cc CommonConfig, opts
 				options = append(options, withTagPromotion(r, branchName))
 			}
 
+			fromImage := "src"
 			srcImageDockerfile, err := discoverSourceImageDockerfile(r)
 			if err != nil {
 				return nil, err
 			}
-			fromImage := "src"
 			if srcImageDockerfile != "" {
 				fromImage = toImage(r, ImageInput{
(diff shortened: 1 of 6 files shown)


mgencur:
/hold

Until tests from https://github.com/openshift/release/pull/44890 pass

pierDipi (approved):
/lgtm
/approve

mgencur:
Hey @pierDipi , I have reworked this a little bit. I think this new approach is better because it will also process image inputs of the source-image and update base_images section. In this way it will be more ready for serverless-operator repo where we have some inputs and base_images.

mgencur:
/unhold
The previous run properly used the source image: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/44890/rehearse-44890-pull-ci-openshift-knative-eventing-release-v1.11-411-test-reconciler-aws-ocp-411/1717467114517827584

mgencur:
Ouch. Some repositories don't have that source-image Dockerfile. I will need to fix it in this PR.

pierDipi (approved):
/lgtm
/approve

PR #85, opened 2023-11-09
Make Serving perf tests run on demand

skonto:
- Let's not run them on every PR by default all the time.
- Verified locally that this generates pre-submits with `always_run: false`.


ReToCode (approved):
/lgtm
/approve

skonto:
/unhold

Right now we don't have pre-submit coverage anyway so we can't test against the onDemand flag.

PR #91, opened 2023-11-15
Add static-images as default Dockerfile includes pattern

pierDipi:
To avoid breaking EKB we need to add a new pattern `openshift/ci-operator/static-images.*` for repositories that want to define Dockerfiles that are not generated by `hack/cmd/generate` (like EKB).


ReToCode (approved):
/lgtm
/approve

PR #92, opened 2023-11-15
Fix onDemand defined for certain OpenShift version

mgencur:
The variant might be e.g. "4.13-ovn" or "411" for auto-generated jobs. We need to remove the dots to match the versions properly.


mgencur:
I have checked that the generated diff is correct: https://github.com/mgencur/release/commit/29115f484e7a6698d10547fc590e20489c8b7057

ReToCode (approved):
/lgtm
/approve

PR #95, opened 2023-11-30
Bump Eventing's OCP to version 4.14 since build pools are now available for it

matzew:
as per title

/cc @pierDipi


pierDipi (approved):
/lgtm
/approve

PR #96, opened 2023-12-08
[WIP] Generate weekly jobs for specific platforms

mgencur:
This will automatically generate jobs for Azure, GCP, vSphere, Hypershift, OVN, OSD, single-node

Individual OpenShift versions can specify `generateCustomConfigs: true`. The repository configuration should then list custom configurations under `customConfigs`. The `releaseBuildConfiguration` key should include at least `tests` key
with the list of tests to be run.

For custom configurations, tests are not generated from Makefile targets but rather taken directly from the configuration. The resulting build configuration is then enriched with images, base images, and dependencies for test steps.


mgencur:
I'm testing the generated configs in https://github.com/openshift/release/pull/46579

config/serverless-operator.yaml
```diff
@@ -71,3 +65,191 @@ repositories:
...
+                        memory: 6Gi
+                      requests:
+                        cpu: 100m
+                        memory: 200Mi
+                workflow: openshift-e2e-gcp
+      - name: "hypershift" # Results in openshift-knative-serverless-operator-<BRANCH>__<OCP_VERSION>-hypershift.yaml
+        releaseBuildConfiguration:
+          base_images:
+            hypershift-operator:
+              name: "4.13"
```
mgencur, line 130: This is one of the few places where the version is hardcoded in the custom config. I'm not sure how to do it better. We could maybe add these base images by default to all configs and have the version generated. It shouldn't harm.

config/serverless-operator.yaml
```diff
@@ -71,3 +65,191 @@ repositories:
...
+              tag: hypershift-operator
+            upi-installer:
+              name: "4.13"
+              namespace: ocp
+              tag: upi-installer
+          releases:
+            latest:
+              integration:
+                include_built_images: true
+                name: "4.13"
```
mgencur, line 141: Another place where we have to specify a version and use "integration" stream due to a bug in CI: https://issues.redhat.com/browse/DPTP-3602
Using "integration" instead of "relaase" is a workaround.

config/serverless-operator.yaml
```diff
@@ -71,3 +65,191 @@ repositories:
...
+      - name: "osd" # Results in openshift-knative-serverless-operator-<BRANCH>__<OCP_VERSION>-osd.yaml
+        releaseBuildConfiguration:
+          tests:
+            - as: e2e-osd-continuous
+              cron: 0 0 * * 0
+              steps:
+                cluster_profile: osd-ephemeral
+                env:
+                  CLUSTER_DURATION: "10800"
+                  CLUSTER_VERSION: "4.13"
```
mgencur, line 194: Another place where the version must be explicitly set.

config/serverless-operator.yaml
mgencur: One possibility would be moving this whole `customConfigs` section under openshift-knative/serverless-operator and have the version properly generated there. But it would complicate the whole workflow - there would be two rounds of generating for the CI config (one in s-o, and then one in hack).
It might be alright to keep it where it is now and update the version in a few places. After all, It's just in one file.

config/serverless-operator.yaml
```diff
@@ -71,3 +66,197 @@ repositories:
...
+              namespace: ci
+              tag: latest
+          tests:
+            - as: e2e-osd-continuous
+              cron: 0 0 * * 0
+              steps:
+                cluster_profile: osd-ephemeral
+                env:
+                  CLUSTER_DURATION: "10800"
+                  CLUSTER_VERSION: "4.13"
```
pierDipi, line 200: If this generates 2 files like 
```
openshift-knative-serverless-operator-main__413-osd.yaml
openshift-knative-serverless-operator-main__411-osd.yaml
```

shouldn't `env.CLUSTER_VERION` be based on the actual osd version ?

mgencur: It is only generated for one version due to this: https://github.com/openshift-knative/hack/pull/96/files#diff-8c104313544b2c9a8413d2cbb7a399e56793ffab4a3e554d5ac2243a5a74f1c8R10
I'm describing the whole issue in https://github.com/openshift-knative/hack/pull/96/files#r1420289826

We'd need to add some kind of generating this.

config/eventing-istio.yaml
```diff
@@ -4,20 +4,20 @@ config:
   branches:
     "release-v1.9":
       openShiftVersions:
-        - version: 4.13
-        - version: 4.10
+        - version: "4.13"
```
skonto, line 7: Is this required?

config/serverless-operator.yaml
pierDipi: Another approach could be to special case serverless-operator in the code itself and inject all these flavor tests for SO only but we would base the various versions on the actual OCP version we're iterating on

mgencur: Another option could be having the custom config as a Go template. On this line above `name: {  .OCPVersion }`. And then we could fill in the template before parsing the configuration.

config/serverless-operator.yaml
```diff
@@ -71,3 +66,197 @@ repositories:
...
+                  COMPUTE_NODES: "4"
+                post:
+                  - chain: gather
+                  - ref: osd-delete-delete
+                pre:
+                  - ref: ipi-install-rbac
+                  - ref: osd-create-create
+                test:
+                  - as: serverless-e2e
+                    commands: make test-e2e-with-kafka test-upstream-e2e-no-upgrade
```
pierDipi, line 210: would be enough to just run `test-e2e-with-kafka` ?

pierDipi: but that's a separate problem, so let's not discuss it here I think

config/serverless-operator.yaml
ReToCode: > Another approach could be to special case serverless-operator in the code itself and inject all these flavor tests for SO only but we would base the various versions on the actual OCP version we're iterating on

I like that suggestion.

config/serverless-operator.yaml
```diff
@@ -71,3 +66,197 @@ repositories:
...
+              namespace: ocp
+              tag: upi-installer
+          tests:
+            - as: e2e-vsphere-continuous
+              cron: 0 0 * * 0
+              steps:
+                cluster_profile: vsphere-2
+                test:
+                  - as: serverless-e2e
+                    commands: make install-mesh test-e2e-with-kafka
```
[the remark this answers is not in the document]
mgencur: It just copies what was there before. Depending on this some tests from servinge2e/kourier in s-o are not run, such as TestKsvcWithServiceMeshSidecar or JWT integration. I don't remember why we enabled mesh for this particular case.

PR #100, opened 2024-01-12
Adjust test suite to cluster size

mgencur:
* vSphere and single-node used to have issues with resources. Reduce the deployment size to give more resources for tests.
* Extend test coverage for platforms where scaling the cluster is possible: GCP, Azure, AWS-OVN
* Reduce test coverage for OSD, Hypershift until https://issues.redhat.com/browse/SRVKS-1187 is fixed.


config/serverless-operator.yaml
```diff
@@ -76,8 +76,28 @@ repositories:
               steps:
                 cluster_profile: aws
                 test:
-                  - as: serverless-e2e
-                    commands: make install-mesh test-e2e-with-kafka
```
maschmid, line 80: do we really want to drop the `install-mesh` ? (are there any optional tests that only run with servicemesh installed we still want to run here?)

mgencur: I'm not sure. Removing this saves some resources.
There are a couple of tests which run with Kourier gateway but together with ServiceMesh as described in [docs](https://docs.openshift.com/serverless/1.31/integrations/serverless-ossm-setup.html#serverless-ossm-setup-with-kourier_serverless-ossm-setup):
TestKsvcWithServiceMeshJWTDefaultPolicy
TestKsvcWithServiceMeshSidecar

This is a pretty minor feature but I could return this back.

maschmid: OK, I agree, let's remove them, FWIW, the feature is deprecated since 1.31 per https://issues.redhat.com/browse/SRVKS-1114

PR #106, opened 2024-01-19
Generate onDemand jobs for next candidate release

### LINKED ISSUE #105: Support clusters with candidate OCP release
The tooling currently generates jobs that use cluster pools. However, it might be required to test with a candidate release that was not included in cluster pools yet (like OCP 4.15 at the moment).
There should be an option to choose starting the cluster from scratch, together with an option to choose the exact release through the `release` stanza, e.g.:
```
releases:
  latest:
    candidate:
      product: ocp
      stream: nightly
      version: "4.15"
```

mgencur:
Fixes #105

"On demand" jobs are generated automatically for OCP version that is equal to MaxVersion+1, where MaxVersion is the highest OpenShift version defined in the yaml config (i.e. when yaml config includes 4.11 and 4.14, the on-demand jobs will be generated for 4.15).
Periodic jobs are NOT generated for this extra version.

### DIFF
--- a/pkg/prowgen/prowgen_config.go
+++ b/pkg/prowgen/prowgen_config.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/coreos/go-semver/semver"
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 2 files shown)


mgencur:
Testing this in https://github.com/openshift/release/pull/47868

pierDipi (approved):
/lgtm
/approve

PR #110, opened 2024-01-24
Add eventing configs for 1.13 branches

creydr:
Adding the configs for eventing-core, eventing-istio and EKB for 1.13 branches.

/hold
until we have branches from upstream

/cc @pierDipi @matzew


creydr:
/unhold
as the 1.13 branches are in place now

/cc @matzew @pierDipi

pierDipi (approved):
/lgtm
/approve

PR #112, opened 2024-01-29
Checkout code before setup go to use go.mod for Go version

pierDipi:
Related to https://github.com/openshift-knative/hack/pull/103 + Bump Go to 1.21

### DIFF
--- a/pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
+++ b/pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
@@ -1,7 +1,7 @@
 # DO NOT EDIT! Generated Dockerfile.
 
 # Dockerfile to bootstrap build and test in openshift-ci
-FROM registry.ci.openshift.org/openshift/release:golang-1.18 as builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.21 as builder
 
 RUN echo "[kubernetes]" >> /etc/yum.repos.d/kubernetes.repo && \
     echo "name=Kubernetes" >> /etc/yum.repos.d/kubernetes.repo && \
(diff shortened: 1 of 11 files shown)


pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # DO NOT EDIT! Generated Dockerfile.
 
 # Dockerfile to bootstrap build and test in openshift-ci
-FROM registry.ci.openshift.org/openshift/release:golang-1.18 as builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.21 as builder
```
mgencur, line 4: Looks good. But when we hit issues like the one from [this thread](https://redhat-internal.slack.com/archives/CD87JDUB0/p1706022237180709) let's use one of the other images. It might be better to do it soon anyway.

mgencur: I could probably test the new images that have ubi-minimal as the base image in other repos (Eventing, Serving, S-O). And then update the Golang base image in a follow-up PR.

pierDipi: Agree, we've tested the new images in Eventing and EKB already so we can change the default here in a follow up PR

pierDipi: See https://github.com/openshift-knative/eventing/commit/e8eb1843a421996968e3062760f7e700baddd9c9

mgencur: Nice!

mgencur: /lgtm

PR #115, opened 2024-02-02
Shorter generated suffixes 

pierDipi:
This will allow having less tests job generated with sha
as they are hard to map to the actual running tests.

Instead of `-continuous`, just use `-c`, and `ocp` is omitted as it's redundant for now.


Cali0707 (approved):
/lgtm

PR #123, opened 2024-02-21
Fix injector to work with all repos under openshift-knative

mgencur:
This is required to properly mark jobs as `always_run: false` when running e.g. with candidate OCP release. So far it was only working for s-o.


pierDipi (approved):
/lgtm
/approve

PR #124, opened 2024-02-22
prowcopy runs injectors configured from source branch

mgencur:
When copying CI config from the source branch and generating the Prow jobs again, it is necessary to apply the same injectors to the generated Prow jobs as for the source branch.

* Removed the `Inject` function on individual injectors because it makes them less versatile (and they're too complex) like in this case where I need to pass the old branch. I left the Inject function only on the group of injectors.

### DIFF
--- a/pkg/prowcopy/prowcopy.go
+++ b/pkg/prowcopy/prowcopy.go
@@ -25,6 +25,7 @@ type Config struct {
 	Tag            string
 	RemovePeriodic bool
 	Remote         string
+	Config         string
 }
 
 // Main is the main function for prowcopy.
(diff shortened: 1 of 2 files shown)


mgencur:
This needs to be followed up by PRs to individual repos that will pass the specific "config" argument, like on this line: https://github.com/openshift-knative/serverless-operator/blob/main/.github/workflows/release-generate-ci.yaml#L63

pkg/prowcopy/prowcopy.go
```diff
@@ -175,3 +208,51 @@ func getJobConfig(match string, c Config) (*prowgen.ReleaseBuildConfiguration, e
...
+	for _, jci := range jcis {
+		sourceBranchName, targetBranch := prowcopyCfg.FromBranch, prowcopyCfg.Branch
+		var sourceBranch *prowgen.Branch
+		// Injectors need to be applied to the new branch in the same way as they were applied
+		// to the source branch when its config was generated.
+		for branchName, branch := range prowgenCfg.Config.Branches {
+			if branchName == sourceBranchName {
+				sourceBranch = &branch
+			}
+		}
```
pierDipi, line 226: Can we just do

```golang
sb, ok := prowgenCfg.Config.Branches[sourceBranchName]
if !ok { }
sourceBranch := &sb
```

pierDipi: though, it's a nit as the list is very small to not warrant the optimization

pierDipi (approved):
I think it's fine to break the workflow until we update the SO GH action with the new flag

nit comment can be addressed later if applicable

pierDipi:
> This needs to be followed up by PRs to individual repos that will pass the specific "config" argument, like on this line: https://github.com/openshift-knative/serverless-operator/blob/main/.github/workflows/release-generate-ci.yaml#L63

AFAIK, Only SO is using this

pkg/prowcopy/prowcopy.go
mgencur: Sorry. Will do, just a moment.

mgencur: Nevermind. Next time.

PR #127, opened 2024-02-28
Prowcopy properly searches for old configs to be deleted

mgencur:
* Also fix the nit from previous PR about getting the right source branch from a map


pierDipi (approved):
Thanks!

/lgtm
/approve

PR #132, opened 2024-03-18
Reduce periodic tests for "released" versions

mgencur:
* Define skipCron flag for OpenShift type

Related [slack discussion](https://redhat-internal.slack.com/archives/CD87JDUB0/p1710763119496039)

### DIFF
--- a/config/eventing-istio.yaml
+++ b/config/eventing-istio.yaml
@@ -7,13 +7,14 @@ config:
         version: "4.12"
     release-v1.10:
       openShiftVersions:
-      - version: "4.14"
-      - onDemand: true
-        version: "4.11"
+      - skipCron: true
+        version: "4.14"
     release-v1.11:
       openShiftVersions:
-      - version: "4.14"
+      - skipCron: true
+        version: "4.14"
       - onDemand: true
+        skipCron: true
         version: "4.12"
     release-v1.12:
       openShiftVersions:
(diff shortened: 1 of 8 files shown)


config/eventing-istio.yaml
```diff
@@ -7,13 +7,14 @@ config:
         version: "4.12"
     release-v1.10:
       openShiftVersions:
-      - version: "4.14"
-      - onDemand: true
```
ReToCode, line 11: was this intentional?

mgencur: Yeah, sorry. Updated the PR description with "Remove OCP 4.11 completely as it's not supported anymore". It was intentional.

pkg/prowgen/prowgen_tests.go
```diff
@@ -166,7 +166,7 @@ func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, sk
 			preSubmitConfiguration.RunIfChanged = test.RunIfChanged
 			cfg.Tests = append(cfg.Tests, *preSubmitConfiguration)
 
-			if !test.SkipCron && !openShift.CandidateRelease {
+			if !test.SkipCron && !openShift.SkipCron && !openShift.CandidateRelease {
```
ReToCode, line 169: this now is a bit confusing, double "skipCron". Can we somehow align this?

mgencur: The `test.SkipCron` is for a single test only. The `openShift.SkipCron` if for all tests running on the given OpenShift version. You might require running periodic job for a given OpenShift version but only skip generating the cron job for a single test.

Would a comment in code make it more clear?

ReToCode: Yeah, I think that explanation was missing for me. If you can explain this in a comment, would be great.

mgencur: Added clarification, please let me know if this is clearer.

PR #134, opened 2024-03-19
Use serverless-ci cluster pool

mgencur:
Depends on https://github.com/openshift/release/pull/49904

Related to https://issues.redhat.com/browse/SRVCOM-3056

* Bump OCP 4.14 to 4.15 as the PR above introduces a cluster pool for 4.15 that will be used
* Run all tests for 4.15 (both periodic and pre-submit) on the existing cluster pool
* Run tests for 4.14 and older by starting a new cluster from scratch, i.e. migrating off the openshift-ci cluster pool. Previously, tests for all versions were running on various cluster pools (4.12 - 4.14) owned by OpenShift team.
* Tests on specific platforms (Azure, vSphere, Hypershift) will be staring clusters from scratch regardless of the OCP version because the existing cluster pool is a simple variant (IPI, AWS).

/hold

### DIFF
--- a/pkg/prowgen/prowgen_tests.go
+++ b/pkg/prowgen/prowgen_tests.go
@@ -24,9 +24,10 @@ const (
 	cronTemplate         = "%d %d * * 2,6"
 	seed                 = 12345
 	devclusterBaseDomain = "serverless.devcluster.openshift.com"
-	// Holds version of the cluster pool dedicated to OpenShift Serverless in CI.
+	// Holds version of the existing cluster pool dedicated to OpenShift Serverless in CI.
 	// See https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools
-	clusterPoolVersion = "4.15"
+	clusterPoolVersion       = "4.15"
+	serverlessClusterProfile = "aws-serverless"
 )
 
 func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, skipE2ETestMatch []string, random *rand.Rand) ReleaseBuildConfigurationOption {
@@ -71,7 +72,7 @@ func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, sk
 				}
 				workflow = pointer.String("generic-claim")
 			} else {
-				clusterProfile = "aws-serverless"
+				clusterProfile = serverlessClusterProfile
 				env = map[string]string{
 					"BASE_DOMAIN": devclusterBaseDomain,
 				}
--- a/pkg/prowgen/prowgen_tests.go
+++ b/pkg/prowgen/prowgen_tests.go
@@ -24,6 +24,9 @@ const (
 	cronTemplate         = "%d %d * * 2,6"
 	seed                 = 12345
 	devclusterBaseDomain = "serverless.devcluster.openshift.com"
+	// Holds version of the cluster pool dedicated to OpenShift Serverless in CI.
+	// See https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools
+	clusterPoolVersion = "4.15"
 )
 
 func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, skipE2ETestMatch []string, random *rand.Rand) ReleaseBuildConfigurationOption {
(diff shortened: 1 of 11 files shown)


mgencur:
Testing the generated configurations in https://github.com/openshift/release/pull/50032

pkg/prowgen/prowgen_tests.go
```diff
@@ -24,6 +24,10 @@ const (
 	cronTemplate         = "%d %d * * 2,6"
 	seed                 = 12345
 	devclusterBaseDomain = "serverless.devcluster.openshift.com"
+	// Holds version of the existing cluster pool dedicated to OpenShift Serverless in CI.
+	// See https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools
+	clusterPoolVersion       = "4.15"
+	serverlessClusterProfile = "aws-serverless"
```
ReToCode, line 30: What does `aws-serverless` refer to?

pkg/prowgen/prowgen_tests.go
```diff
@@ -48,27 +52,29 @@ func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, sk
...
+
+			useClusterPool := openShift.Version == clusterPoolVersion
+			// Make sure to use the existing cluster pool if available for the given OpenShift version.
+			if useClusterPool {
+				clusterClaim = &cioperatorapi.ClusterClaim{
+					Product:      cioperatorapi.ReleaseProductOCP,
+					Version:      openShift.Version,
+					Architecture: cioperatorapi.ReleaseArchitectureAMD64,
+					Cloud:        cioperatorapi.CloudAWS,
+					Owner:        "serverless-ci",
```
ReToCode, line 70: might be worth to add a comment here that links to the place in openshift/release that defines `serverless-ci`. Just for later people to understand the link.

pkg/prowgen/prowgen_tests.go
mgencur: It's the name of the profile that we've been using for starting new clusters from scratch for pre-released versions of OCP (so far it was 4.15, now it will be 4.16).
It's configured [here](https://github.com/openshift/ci-tools/pull/3978/files#diff-dfee74d02fd350fb1ee1ffbbbc32acf4cb075c9ae42b6e31c8c7acefe7fed623R1237)
In the current PR, I'm only creating a new constant for this. Previously, it was used directly in code [here](https://github.com/openshift-knative/hack/blob/main/pkg/prowgen/prowgen_tests.go#L68)

pkg/prowgen/prowgen_tests.go
```diff
@@ -48,27 +52,29 @@ func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, sk
...
+				clusterClaim   *cioperatorapi.ClusterClaim
+				clusterProfile cioperatorapi.ClusterProfile
+				workflow       *string
+				env            cioperatorapi.TestEnvironment
+			)
+
+			useClusterPool := openShift.Version == clusterPoolVersion
+			// Make sure to use the existing cluster pool if available for the given OpenShift version.
+			if useClusterPool {
+				clusterClaim = &cioperatorapi.ClusterClaim{
```
ReToCode, line 65: urgh, that configuration is hard to parse (cause of prow I assume). Let me see if I get this right:
- if we want to use our pool, we set `clusterClaim` otherwise it is `nil`
- if we do not want to use our pool, we set `clusterProfile` to `aws-serverless` otherwise it is nil.
- we have to set workflow in both cases

I think same question as above, why do we refer to `aws-serverless` here when we are not using our pool. What does that define?

pkg/prowgen/prowgen_tests.go
ReToCode: Can we maybe link that as well (see other comment below).

pkg/prowgen/prowgen_tests.go
mgencur: Yeah, so I will update this PR after https://github.com/openshift/release/pull/49904 is merged so that I can directly point to where it's defined in the codebase.

pkg/prowgen/prowgen_tests.go
mgencur: You're correct.
* For cluster pools, we need to use `clusterClaim` and the workflow needs to be `generic-claim`
* For clusters starting from scratch, we need to use `clusterProfile` (this is mutually exclusive with `clusterClaim`) and the workflow needs to be `ipi-aws`
Regarding `aws-serverless`, this is answered in the other comment.

mgencur: If starting a cluster from scratch, we also need to reference a "post" step `ipi-deprovision-deprovision`

ReToCode (approved):
/lgtm

PR #138, opened 2024-03-28
Fix OCP version for hypefoil repo and skip cron

pierDipi:
- We cannot yet use 4.15 as AMQ Stream doesn't support it yet
- Remove periodic and pre-submite run for hypefoil tests (to be intentional at running those tests when needed as they are heavy)
- Improve match regex


ReToCode (approved):
/lgtm
/approve

PR #144, opened 2024-04-03
OWNERS_ALIASES unsupported on openshift/release subdirectories

pierDipi:
See https://github.com/openshift/release/pull/50459#issuecomment-2033695718


aliok (approved):
/lgtm
/approve

I am ok with always having the same folks :)

PR #146, opened 2024-04-03
Build images in pre-submit phase only for mandatory OpenShift versions

mgencur:
This is to reduce image builds in pre-submit phase. Building other variants such as GCP, Azure, OSD, or older OCP versions is not necessary.

This will generate changes like in a test PR: https://github.com/openshift/release/pull/50537


pierDipi (approved):
Nice!

/lgtm
/approve

PR #148, opened 2024-04-25
Add new midstream branches

serverless-qe:
Discover new branches using openshift-knative/hack.


ReToCode:
We don't want 1.13 CI because these branches are not needed. @mgencur I expect this is going to be recreated when I close it? @skonto so we might need to rename the branches?

pierDipi:
> @mgencur I expect this is going to be recreated when I close it?

Yes, @ReToCode

ReToCode:
I did rename the branches, so this should be fine with closing.

PR #77, opened 2023-09-29
Skip mesh auth policy gen in tests

skonto:
Follow up from here: https://github.com/openshift-knative/serverless-operator/pull/2290

PR #154, opened 2024-05-09
SRVCLI-401: :gift: Add kn-event configuration

cardil:
## Changes

 - :gift: Add kn-event configuration

Fixes [SRVCLI-401](https://issues.redhat.com/browse/SRVCLI-401)

Here's the current config diff for kn-event: https://gist.github.com/cardil/adedaf1313446a6329fb9a958670198a (I'll keep it updated)


cardil:
Require https://github.com/openshift-knative/kn-plugin-event/pull/401

config/kn-event.yaml
```diff
@@ -0,0 +1,50 @@
...
+    release-1.14:
+      openShiftVersions:
+        - version: '4.15'
+    release-next:
+      openShiftVersions:
+        - version: '4.15'
+repositories:
+  - org: openshift-knative
+    repo: kn-plugin-event
+    imagePrefix: client-plugin-event
```
dsimansk, line 20: It's a bit nit picking, but I wonder how the final names is generated from all those values. Is it going too verbose? I'd keep `kn-plugin-event` here. Wdyt?

config/kn-event.yaml
```diff
@@ -0,0 +1,50 @@
...
+    binaryBuildCommands: |-
+      export HOME=/tmp
+      export TEST_IMAGES_DIR=/usr/bin
+      ./mage build
```
dsimansk, line 24: In case of client, I've been looking at this stanza and my thought's been to wrap it into Dockerfile stage build to simplify CI config. Similarly to `Makefile` PR in the kn-plugin-event midstream, but also execute the make withing Docker builds.

config/serverless-operator.yaml
```diff
@@ -297,30 +297,46 @@ repositories:
     - olm-catalog/serverless-operator/index/Dockerfile
   e2e:
   - match: ^operator-e2e$
+    env:
+      SKIP_MESH_AUTH_POLICY_GENERATION: "true"
```
dsimansk, line 301: Is it related to work on event plugin? Would it make more sense to check in S-O Makefile `if PROW` and based on that define defaults skips? Maybe we will need more skips in the future.

Makefile
```diff
@@ -2,29 +2,33 @@
...
 
-generate-ci: clean generate-eventing-ci generate-serving-ci
-.PHONY: generate-ci
+all: eventing serving operator client
+.PHONY: all
 
-generate-ci-no-clean: generate-eventing-ci generate-serving-ci generate-serverless-operator-ci
-.PHONY: generate-ci-no-clean
-
-generate-eventing-ci:
```
dsimansk, line 11: These cmd name changes will most likely break GitHub actions that execute generators periodically.

config/kn-event.yaml
cardil: This is how Openshift CI advises building binaries in a separate step, and later use the `bin` image contents, while building the images.

Since then, most Serverless repos moved to pure Docker approach, when the `bin` building doesn't matter, as everything is done in the build image step: https://github.com/openshift-knative/serving/blob/release-next/openshift/ci-operator/knative-images/activator/Dockerfile

IDK, if I keep kn-event current setup or shift it to how other repos does this. I'm leaning towards the pure Dockerfile approach.

config/serverless-operator.yaml
cardil: This is a rather messy env var introduced in https://github.com/openshift-knative/hack/pull/77

Now, it is added to every command. I didn't like that, so I added support for defining the environmental variables.

I believe this variable makes sense only in terms of S-O, that's why I added it just to its config. Is it right, @skonto?

Makefile
cardil: I'll update the workflows in this repo. Are you aware of other cross-repo references to those targets?

config/kn-event.yaml
cardil: I updated the diff above. I think it's acceptable to proceed.

The naming thing. Yes, we could proceed with an image prefix of `kn-plugin-event`, however:

 - historically, the images were published under the prefix `client-plugin-event`, so keeping it makes the history consistent
 - we already burned down the naming convention in https://github.com/openshift-knative/kn-plugin-event/pull/442 (and its backports)

So, we can change the image prefix to `kn-plugin-event`, but I'm a bit reluctant due to the above points.

config/kn-event.yaml
cardil: I changed my mind to keep the current approach, as I managed to change the code to support both ways. Also, the current kn-event's approach may be better for multi-arch builds in Openshift CI.

Still, maybe, I'll follow up the kn-event to add a Makefile target that detects Openshift CI environment and sets proper environment variables.

Makefile
cardil: The actions are updated, and I restored those targets, as deprecated (they print a warn box when used).

cardil:
/hold

Let me provide the full diff, for all the components...

Makefile
dsimansk: I'd defer name change to @mgencur or @pierDipi to chime in here. 
I'm not a huge fan to carry this change within "add config" PR, because it doesn't seem to me as necessary for adding event plugin config.

pkg/prowgen/prowgen_config.go
```diff
@@ -19,12 +19,15 @@ type Repository struct {
...
 	ImagePrefix           string                                                      `json:"imagePrefix,omitempty" yaml:"imagePrefix,omitempty"`
 	ImageNameOverrides    map[string]string                                           `json:"imageNameOverrides,omitempty" yaml:"imageNameOverrides,omitempty"`
 	SlackChannel          string                                                      `json:"slackChannel,omitempty" yaml:"slackChannel,omitempty"`
 	CanonicalGoRepository *string                                                     `json:"canonicalGoRepository,omitempty" yaml:"canonicalGoRepository,omitempty"`
 	E2ETests              []E2ETest                                                   `json:"e2e,omitempty" yaml:"e2e,omitempty"`
 	Dockerfiles           Dockerfiles                                                 `json:"dockerfiles,omitempty" yaml:"dockerfiles,omitempty"`
+	BaseImages            map[string]cioperatorapi.ImageStreamTagReference            `json:"baseImages,omitempty" yaml:"baseImages,omitempty"`
+	SharedInputs          map[string]cioperatorapi.ImageBuildInputs                   `json:"sharedInputs,omitempty" yaml:"sharedInputs,omitempty"`
```
dsimansk, line 30: Wouldn't it make sense to keep variable name as per CI operator API, `ImageBuildInputs`?

pkg/prowgen/prowgen_config.go
```diff
@@ -349,22 +355,39 @@ func withTagPromotion(r Repository, branchName string) ReleaseBuildConfiguration
...
+	if r.Promotion.OmitSource {
+		return nil
+	}
+	return map[string]string{
+		// Add source image
+		transformLegacyKnativeSourceImageName(r): "src",
+	}
+}
+
+func createPropotionName(p Promotion, branchName string) string {
```
dsimansk, line 379:
```suggestion
func createPromotionName(p Promotion, branchName string) string {
```

A typo, right?

pkg/prowgen/prowgen_images_discovery.go
```diff
@@ -36,16 +37,23 @@ var (
...
-type orgRepoTag struct {
-	Org  string
-	Repo string
-	Tag  string
+type ociReference struct {
+	Domain string
+	Org    string
+	Repo   string
+	Tag    string
+	Digest string
```
dsimansk, line 53: Why is digest needed here?

cardil: We don't use it currently, which doesn't mean at any point such image reference might be used in one of Dockerfiles.

pkg/prowgen/prowgen_config.go
cardil: Yes, I can name it `ImageBuildInputs` although the shared aspect differs here from CI operator API.

In CI operator API, the `ImageBuildInputs` is used for a specific image. Here, I'm using that do add common, provided upfront inputs for all images from a given repo.

I could name it `SharedImageBuildInputs`, but isn't that too mouthful?

cardil:
/close

in favor of https://github.com/openshift-knative/hack/pull/388

PR #155, opened 2024-05-10
SRVCLI-401: :gift_heart: Donate deviate tool

cardil:
Donating the deviate tool from https://github.com/cardil/deviate

Related to [SRVCLI-401](https://issues.redhat.com/browse/SRVCLI-401)


dsimansk:
Would it make sense to make it a submodule isolate related `go.mod` changes?

cardil:
> @dsimansk: Would it make sense to make it a submodule isolate related `go.mod` changes?

Yes, I was thinking the same. My initial idea was to change the repo's layout to mirror the https://github.com/knative/toolbox (and rename the repo as well). But, the team was reluctant to rename because of short terms breakages. The same applies to any re-layout. Personally, I would go for it, as the breakages aren't a big problem here, and could be quickly fixed.

cardil:
/close

In favor of https://github.com/openshift-knative/deviate

PR #157, opened 2024-05-23
Skip 1.12 in SO branch name calculation

creydr:
As we kind of skipped the upstream 1.12 branch name (1.12 was actually upstream 1.13), we should reflect this in the SO branch name generation.

So we have:
* 1.12 (& 1.13) -> 1.33
* 1.14 -> 1.34


ReToCode (approved):
/lgtm
/approve

PR #161, opened 2024-06-14
Generate Konflux configurations

pierDipi:
Integrate KonfluxGen with prowgen so that after running prowgen we can create PRs for all midstream repositories to add/update/remove Konflux components and applications (to be added pipelines in `.tekton`).

The idea is:

1. Configure branch to enable Konflux generation `konflux: enabled: true`
2. If enabled create a PR to create/update/remove components/applications for each configured branch and repository (and as a follow up pipelines and ImageRepository)

To create PRs we need to generate one step per repository per branch and so there is a new command `make update-action` and the consistency is verified as part of the `make unit-tests` command

### DIFF
--- a/.github/workflows/release-generate-ci-template.yaml
+++ b/.github/workflows/release-generate-ci-template.yaml
@@ -1,0 +1,80 @@
+name: Generate CI config (template)
+
+on:
+  pull_request:
+    branches:
+      - '**'
+  push:
+    branches:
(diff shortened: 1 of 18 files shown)


mgencur:
LGTM.
@pierDipi Do you want to merge it now?

pierDipi:
I would merge and try it !

PR #162, opened 2024-06-18
Add new midstream branches

serverless-qe:
Discover new branches using openshift-knative/hack.

### DIFF
--- a/config/backstage-plugins.yaml
+++ b/config/backstage-plugins.yaml
@@ -1,6 +1,9 @@
 config:
   branches:
     release-v1.12:
+      konflux:
+        Enabled: false
+        Nudges: null
       openShiftVersions:
       - version: "4.15"
       - onDemand: true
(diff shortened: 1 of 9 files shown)


config/backstage-plugins.yaml
```diff
@@ -1,6 +1,9 @@
 config:
   branches:
     release-v1.12:
+      konflux:
```
mgencur, line 4: I think it might have been better to define ``json:"enabled,omitempty" yaml:"nudges,omitempty"`` here? https://github.com/openshift-knative/hack/pull/161/files#diff-9cb0409bd8a2445656e16ee55c0cc76205f9e9753d4cafa7d59d9bd879395ffdR78

mgencur: @pierDipi

pierDipi: Yes, let me fix that

pierDipi: https://github.com/openshift-knative/hack/pull/164

PR #163, opened 2024-06-18
Generate Konflux pipelines

pierDipi:
Follow up to https://github.com/openshift-knative/hack/pull/161

- Generate pipeline and pipeline runs for push and PR referencing the Konflux Docker build pipeline (downloaded from https://github.com/konflux-ci/build-definitions/blob/main/pipelines/docker-build-oci-ta/README.md, see updated cmd/konflux-gen/README.md in the PR)

### DIFF
--- a/pkg/konfluxgen/docker-build-oci-ta.yaml
+++ b/pkg/konfluxgen/docker-build-oci-ta.yaml
@@ -1,0 +1,311 @@
+apiVersion: tekton.dev/v1
+kind: Pipeline
+metadata:
+  labels:
+    pipelines.openshift.io/runtime: generic
+    pipelines.openshift.io/strategy: docker
+    pipelines.openshift.io/used-by: build-cloud
+  name: docker-build-oci-ta
+spec:
+  finally:
+  - name: show-sbom
+    params:
+    - name: IMAGE_URL
+      value: $(tasks.build-container.results.IMAGE_URL)
+    taskRef:
+      name: show-sbom
+      version: "0.1"
+  params:
+  - description: Source Repository URL
+    name: git-url
+    type: string
+  - default: ""
+    description: Revision of the Source Repository
+    name: revision
+    type: string
+  - description: Fully Qualified Output Image
+    name: output-image
+    type: string
+  - default: .
+    description: Path to the source code of an application's component from where
+      to build image.
+    name: path-context
+    type: string
+  - default: Dockerfile
+    description: Path to the Dockerfile inside the context specified by parameter
+      path-context
+    name: dockerfile
+    type: string
+  - default: "false"
+    description: Force rebuild image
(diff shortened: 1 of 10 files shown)


mgencur:
I haven't fully reviewed this but I can see there's just one type of Pipeline. I was working on https://github.com/openshift-knative/serverless-operator/pull/2707 to build all images there and figured  there are two Pipelines that are slightly different. I call them `serverless-image` and `serverless-index` (the Pipeline for the index image is slightly different). Maybe @creydr could shed some light on the difference as I just copied them from the original pipelines. 
The index image is probably not considered in this PR. Just something to thing about for the future.

creydr:
> I haven't fully reviewed this but I can see there's just one type of Pipeline. I was working on [openshift-knative/serverless-operator#2707](https://github.com/openshift-knative/serverless-operator/pull/2707) to build all images there and figured there are two Pipelines that are slightly different. I call them `serverless-image` and `serverless-index` (the Pipeline for the index image is slightly different). Maybe @creydr could shed some light on the difference as I just copied them from the original pipelines. The index image is probably not considered in this PR. Just something to think about for the future.

The serverless-index pipeline is a bit different, because it has additional steps to verify the file based catalog of the index. This was mostly auto generated, as Konflux recognized it as an index build (therefor I needed to reference a devfile, which I reverted, after we had the pipeline generated (https://github.com/openshift-knative/serverless-operator/pull/2674/commits/2a6573d20e009866cc7b31f094393b1dae6b4b53), as this made things more complicated and was not needed anymore (as we had the FBC related tasks in the pipeline)).

pierDipi:
That pipeline will need to be special cases, it's this one basically https://github.com/konflux-ci/build-definitions/tree/main/pipelines/fbc-builder

pierDipi:
But for now, I'm focusing on dependents component, SO will come next

creydr:
> But for now, I'm focusing on dependents component, SO will come next

ACK. Just wanted to mention it in this context, so we don't forget about it

pierDipi:
Let me know if the PR looks ok and code is clean/readable enough as I'd like to try it out for eventing core

pkg/konfluxgen/docker-build-oci-ta.yaml
creydr, line 1: Only to make sure I understand this correctly: this is the "default-custom" pipeline from Konflux?

pierDipi: Yes

pierDipi: Actually, it's the new one that is not using PVC for storing artifacts and should solve our issue with Quota

creydr (commented):
LGTM from my side

/lgtm

PR #166, opened 2024-06-18
Fix midstream repository directory for PR

pierDipi:


pierDipi:
/cc @creydr @mgencur 

A missing "dot"

PR #167, opened 2024-06-18
Prowgen: support passing directory to make the generation faster

pierDipi:
"Generate CI" step takes between 20 to 30 minutes to run. (https://github.com/openshift-knative/hack/actions/runs/9567605982/job/26375754744)

This should speed up the `make generate-ci` command as we won't run `make ci-operator-config jobs` multiple times (twice per file vs just twice)

Edit: now it takes only 5 minutes :tada:  https://github.com/openshift-knative/hack/actions/runs/9569059982/job/26380732319?pr=167


creydr (approved):
/lgtm

PR #168, opened 2024-06-19
Fix create PR for konflux configurations

pierDipi:
Use gh CLI as the create-pull-request action doesn't work with existing branches https://github.com/peter-evans/create-pull-request/issues/2108


creydr (approved):
/lgtm

PR #169, opened 2024-06-19
Migrate to use GH CLI even for openshift/release PR

pierDipi:
I don't know why but it's failing and for consistency using the same method is better https://github.com/openshift-knative/hack/actions/runs/9583203242/job/26423940658


creydr (approved):
/lgtm

PR #172, opened 2024-06-20
Use x-access-token

pierDipi:
Trying to fix https://github.com/openshift-knative/hack/actions/runs/9594204030 using this approach https://stackoverflow.com/questions/57921401/push-to-origin-from-github-action


pierDipi:
I will try ...

PR #173, opened 2024-06-25
Add serverless-qe when adding remote and default bash shell

pierDipi:


.github/workflows/release-generate-ci-template.yaml
```diff
@@ -68,5 +72,5 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.SERVERLESS_QE_ROBOT }}
           GITHUB_TOKEN: ${{ secrets.SERVERLESS_QE_ROBOT }}
-        run: git remote add fork https://${{ secrets.SERVERLESS_QE_ROBOT }}@github.com/serverless-qe/release.git && git push fork sync-serverless-ci:sync-serverless-ci -f && gh pr create --base master --head serverless-qe:sync-serverless-ci --fill-verbose
+        run: git remote add fork https://serverless-qe:${{ secrets.SERVERLESS_QE_ROBOT }}@github.com/serverless-qe/release.git && git push fork sync-serverless-ci:sync-serverless-ci -f && gh pr create --base master --head serverless-qe:sync-serverless-ci --fill-verbose
```
mgencur, line 75:
```suggestion
        run: git remote add fork "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/release.git" && git push fork sync-serverless-ci:sync-serverless-ci -f && gh pr create --base master --head serverless-qe:sync-serverless-ci --fill-verbose
```

mgencur: I would really use simple `${}` and reference the env variable instead of directly the global secret here. And also quote the string to prevent special characters in the token to break shell, like in the example above.

pierDipi:
:crossed_fingers:

PR #177, opened 2024-06-27
Include branch name in Konflux components `metadata.name`

creydr:
Including the components branch name into its metadata.name to be able to have the same component from different branches in different applications linked


pierDipi (approved):
/lgtm
/approve

PR #179, opened 2024-06-27
Fix application name annotation in PipelineRun template

creydr:
`.ApplicationName` has already the branch name


pierDipi (approved):
/lgtm
/approve

PR #181, opened 2024-06-28
Add branch name to pipelinerun filename to align with component name

creydr:
Seems konflux wants the full component name included in the pipelinerun filename and also on the pipelinerun `.metadata.name`. E.g. https://github.com/openshift-knative/eventing/pull/681/files

### DIFF
--- a/pkg/konfluxgen/konfluxgen.go
+++ b/pkg/konfluxgen/konfluxgen.go
@@ -50,6 +50,10 @@ func Generate(cfg Config) error {
 		return fmt.Errorf("failed to remove %q directory: %w", cfg.ResourcesOutputPath, err)
 	}
 
+	if err := os.RemoveAll(cfg.PipelinesOutputPath); err != nil {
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
/lgtm

pkg/konfluxgen/konfluxgen.go
```diff
@@ -50,6 +50,10 @@ func Generate(cfg Config) error {
 		return fmt.Errorf("failed to remove %q directory: %w", cfg.ResourcesOutputPath, err)
 	}
 
+	if err := os.RemoveAll(cfg.PipelinesOutputPath); err != nil {
+		return fmt.Errorf("failed to remove %q directory: %w", cfg.PipelinesOutputPath, err)
+	}
```
pierDipi, line 55: Eventually, I believe we would leverage konflux auto upgrade of pipeline tasks, so I believe we would somehow "ignore changes"

PR #182, opened 2024-07-02
Update konflux pipeline to have bundled tasks

pierDipi:
Konflux uses bundled resolver, so we need to include the tasks with the bundle


creydr (commented):
/lgtm

pierDipi:
@creydr I pushed a temporary fix

PR #204, opened 2024-08-08
Remove unneeded annotations on components

creydr:
As per the discussions in [Slack](https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1723023194521589), the `build.appstudio.openshift.io/request: configure-pac` annotation is allowed only to applied once.

Therefor we remove the unneeded annotations and add an hint in the README.md that this is required to be done once per repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,34 @@ Limitations:
     ```
 3. Run `make konflux-apply`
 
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
/lgtm
/approve

PR #206, opened 2024-08-08
Preserve newer Tekton task images in Konflux pipelines

pierDipi:
This handles the case task images are updated with renovate, like in https://github.com/openshift-knative/eventing-istio/pull/273

### DIFF
--- a/pkg/konfluxgen/konfluxgen.go
+++ b/pkg/konfluxgen/konfluxgen.go
@@ -151,7 +151,7 @@ func Generate(cfg Config) error {
 			if err := applicationTemplate.Execute(buf, config); err != nil {
 				return fmt.Errorf("failed to execute template for application %q: %w", appKey, err)
 			}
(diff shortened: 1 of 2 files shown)


creydr (approved):
/lgtm

creydr:
This is great. Thanks a lot!

pierDipi:
One further issue is that https://github.com/openshift-knative/eventing-istio/pull/273 updates on release-next will not be preserved but that's a bit the nature of release next :(

creydr:
I though about excluding the main branch from this logic, and running the updating the pipeline definition file (https://github.com/openshift-knative/hack/blob/main/pkg/konfluxgen/docker-build.yaml) each time. So we have the latest tags on main / release-next, but ignore them on the release branches.
To not run the renovate jobs on release-next (like https://github.com/openshift-knative/eventing-istio/pull/273), we could check to configure a renovate.json and disable it for this branch

PR #215, opened 2024-08-29
Fix preserve newer task images

pierDipi:
Fix for unwanted PRs https://github.com/openshift-knative/eventing/pull/774

### DIFF
--- a/pkg/konfluxgen/konfluxgen.go
+++ b/pkg/konfluxgen/konfluxgen.go
@@ -394,9 +394,6 @@ func makeValidName(n string) string {
 }
 
 func WriteFileReplacingNewerTaskImages(name string, data []byte, perm os.FileMode) error {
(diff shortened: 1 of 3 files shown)


pkg/konfluxgen/konfluxgen.go
```diff
@@ -394,9 +394,6 @@ func makeValidName(n string) string {
 }
 
 func WriteFileReplacingNewerTaskImages(name string, data []byte, perm os.FileMode) error {
-	if !strings.HasSuffix(name, ".yaml") || !strings.HasSuffix(name, ".yml") {
-		return os.WriteFile(name, data, perm)
-	}
```
pierDipi, line 399: This is the root case but it's also not necessary anymore as I initially was using a YAML-only approach but after I made it file type agnostic

creydr (commented):
/lgtm

PR #222, opened 2024-08-30
Don't delete docker-build and FBC pipeline base

creydr:
Fixes the current issue, that the image tags in the docker-build.yaml and FBC are not kept.

Alternatively, we could also skip the `RemoveAll` on the `.tekton` folder completely. Anyhow this would mean, that we would have issues, when we change the filename pattern :shrug:


pierDipi (approved):
/lgtm
/approve

PR #226, opened 2024-09-02
Tag images with 'latest' and release tags

pierDipi:
This will help with generating the CSV file with the "latest" image SHAs

Here is how it will be used https://github.com/openshift-knative/serverless-operator/pull/2826

### DIFF
--- a/pkg/konfluxgen/konfluxgen.go
+++ b/pkg/konfluxgen/konfluxgen.go
@@ -172,7 +172,7 @@ func Generate(cfg Config) error {
 				}
 			}
 
-			applications[appKey][dockerfileComponentKey(cfg.ComponentNameFunc, c.ReleaseBuildConfiguration, ib)] = DockerfileApplicationConfig{
+			r := DockerfileApplicationConfig{
 				ApplicationName:           cfg.ApplicationName,
 				ComponentName:             Truncate(Sanitize(cfg.ComponentNameFunc(c.ReleaseBuildConfiguration, ib))),
 				ReleaseBuildConfiguration: c.ReleaseBuildConfiguration,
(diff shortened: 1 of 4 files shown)


pkg/konfluxgen/konfluxgen.go
```diff
@@ -181,7 +181,9 @@ func Generate(cfg Config) error {
 				Nudges:                        append(cfg.Nudges, cfg.NudgesFunc(c.ReleaseBuildConfiguration, ib)...),
 				Pipeline:                      pipeline,
 				AdditionalTektonCELExpression: cfg.AdditionalTektonCELExpressionFunc(c.ReleaseBuildConfiguration, ib),
+				Tags:                          []string{"latest"},
```
creydr, line 184: I probably miss something, but just to be sure here: as we use the same pipeline for on-push and on-pr, wouldn't we tag images from PRs also with `latest`?

pierDipi: no, the parameter is only added for `push` pipeline

pierDipi:
```
    {{{- if eq .Event "push" }}}
    - name: additional-tags
      value: [{{{ range $tag := .Tags }}} {{{ $tag }}}, {{{ end }}}]
    {{{- end }}}
```

pierDipi: we're injecting parameters for the pipeline run and we have one pipeline run per event (push and PR separate)

creydr: Ah nice :+1:

creydr (approved):
/lgtm

PR #231, opened 2024-09-03
Fix syntax to pass additional-tags

pierDipi:
Tested here https://github.com/openshift-knative/eventing/pull/779/commits/08bcc977d0f42fa2730aa12dc17836afeee9d54c


creydr (commented):
Seen in https://github.com/openshift-knative/eventing/pull/779

/lgtm

PR #219, opened 2024-08-29
[main] Add Konflux configurations

serverless-qe:
Add Konflux components and pipelines

PR #233, opened 2024-09-03
Fix ImageRepository name and component for SO

pierDipi:
Noticed here https://github.com/openshift-knative/hack/pull/219

Tested that other components produce the same name as currently


creydr (commented):
/lgtm

PR #234, opened 2024-09-03
Fix konflux apply Go setup

pierDipi:
See error https://github.com/openshift-knative/hack/actions/runs/10681881639/job/29606464642

### DIFF
--- a/.github/workflows/apply-konflux-manifests.yaml
+++ b/.github/workflows/apply-konflux-manifests.yaml
@@ -10,14 +10,14 @@ jobs:
   konflux-apply:
     name: Apply Konflux Manifests
     runs-on: ubuntu-latest
-    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/apply-konflux-manifests.yaml
```diff
@@ -10,14 +10,14 @@ jobs:
...
-      GOPATH: ${{ github.workspace }}
     steps:
-      - name: Checkout openshift-knative/hack
+      - name: Check out code
         uses: actions/checkout@v4
 
       - name: Setup Golang
         uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
```
pierDipi, line 20: Copied from go-unit-tests

creydr (commented):
/lgtm

PR #237, opened 2024-09-04
Remove deprecated sbom-json-check step from pipelines

pierDipi:
See https://github.com/konflux-ci/build-definitions/blob/main/task/sbom-json-check/0.2/MIGRATION.md


creydr (approved):
/lgtm

PR #239, opened 2024-09-04
Automated pipelines update

pierDipi:
This will keep pipelines up to date

We need to inject parameters and for that we can use Kustomizations

### DIFF
--- a/.github/workflows/release-discover-branches.yaml
+++ b/.github/workflows/release-discover-branches.yaml
@@ -19,10 +19,6 @@ jobs:
     env:
       GOPATH: ${{ github.workspace }}
     steps:
-      - uses: tektoncd/actions/setup-tektoncd-cli@main
-        with:
-          version: latest
-
       - name: Checkout openshift-knative/hack
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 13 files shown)


.github/workflows/release-discover-branches.yaml
```diff
@@ -30,6 +29,13 @@ jobs:
         with:
           go-version-file: ./src/github.com/openshift-knative/hack/go.mod
 
+      - uses: tektoncd/actions/setup-tektoncd-cli@main
+        with:
+          version: latest
+
+      - name: Install Kustomize
+        run: GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@latest
```
creydr, line 37: nit: kustomize is already in ubuntu-latest runner image installed (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#tools)

pierDipi: I think having a specific version helps?

creydr (commented):
nice. Only one nit. But feel free to unhold it otherwise

/lgtm
/hold

pierDipi:
/unhold

I think it's fine for now

PR #241, opened 2024-09-04
generate: Add possibility to add symlinks pointing to binary

creydr:
Adds the `sym-link-names` arg, which allows to specify symbolic link names which should point to the built binary in the image.
This can be helpful for migrating the kn-plugin-func utils [Dockerfile](https://github.com/knative/func/blob/a13056e4ee5b7e9d9571a97a9867292ac5c7f71b/Dockerfile.utils#L23-L25), which has symlinks.

### DIFF
--- a/cmd/generate/Dockerfile.template
+++ b/cmd/generate/Dockerfile.template
@@ -21,9 +21,9 @@ USER 65532
 
 COPY --from=builder /usr/bin/main /usr/bin/main
 
-{{- if .sym_links }}
+{{- if .post_build_instructions }}
 
-CMD {{ .sym_links }}
+{{ .post_build_instructions }}
 {{- end }}
 
 LABEL \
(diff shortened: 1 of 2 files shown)


cmd/generate/Dockerfile.template
```diff
@@ -21,6 +21,11 @@ USER 65532
 
 COPY --from=builder /usr/bin/main /usr/bin/main
 
+{{- if .sym_links }}
+
+CMD {{ .sym_links }}
+{{- end }}
```
pierDipi, line 27: func-utils is using `RUN`, I would also just call this `.postBuildInstructions` and remove the CMD or RUN

```
RUN ln -s /usr/local/bin/func-util /usr/local/bin/deploy && \
    ln -s /usr/local/bin/func-util /usr/local/bin/scaffold && \
    ln -s /usr/local/bin/func-util /usr/local/bin/s2i
```

{{ .postBuildInstructions }}

creydr: Good point about CMD vs RUN

creydr: Also switched to the postbuildinstructions

pierDipi (approved):
We can always allow more flags to add more instructions if necessary

/lgtm
/approve

PR #242, opened 2024-09-05
Add specific version tag

pierDipi:
Tag images with the specific version tag.

This helps with creating the `ReleasePlanAdmission` for releases.


creydr (commented):
/lgtm

PR #243, opened 2024-09-05
Provide template for target binary name in Dockerfile generator

mgencur:
Some upstream tests from Serving depend on exact path that is compatible with binaries generated by `ko`.
Examples:
https://github.com/knative/serving/blob/main/test/conformance/runtime/cmd_args_test.go#L43
https://github.com/knative/serving/blob/main/test/conformance/runtime/readiness_probe_test.go#L98

This is required to onboard Serving otherwise some tests are failing:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_serving/832/pull-ci-openshift-knative-serving-release-v1.15-415-test-e2e-aws-415/1831386928168046592


pierDipi (approved):
/lgtm
/approve

pierDipi:
I think since this makes bash scripts easier, let's just do it even though it can be achieved differently

PR #244, opened 2024-09-05
Use kubectl and oc from ocp tools image

creydr:
Saw this at [func](https://github.com/openshift-knative/kn-plugin-func/blob/f9f4ecd433329fed9cef7c3f298ef1df1faaf3c9/openshift/ci-operator/build-image/Dockerfile#L3) and liked the idea using it from the tools image (also provides oc - which oc seems to use in some tests)

@mgencur @pierDipi: opinions on using this?

### DIFF
--- a/cmd/generate/dockerfile-templates/BuildImageDockerfile.template
+++ b/cmd/generate/dockerfile-templates/BuildImageDockerfile.template
@@ -1,6 +1,6 @@
 # DO NOT EDIT! Generated Dockerfile.
 
-FROM registry.ci.openshift.org/ocp/4.17:tools as tools
+FROM {{.ocpToolsImage}} as tools
 
 # Dockerfile to bootstrap build and test in openshift-ci
 FROM {{.builder}} as builder
--- a/cmd/generate/dockerfile-templates/BuildImageDockerfile.template
+++ b/cmd/generate/dockerfile-templates/BuildImageDockerfile.template
@@ -1,11 +1,15 @@
 # DO NOT EDIT! Generated Dockerfile.
 
-FROM {{.ocpToolsImage}} as tools
+FROM registry.ci.openshift.org/ocp/4.17:cli-artifacts as tools
 
 # Dockerfile to bootstrap build and test in openshift-ci
 FROM {{.builder}} as builder
 
-COPY --from=tools /usr/bin/oc /usr/bin/kubectl /usr/bin/
+ARG TARGETARCH
+
+COPY --from=tools /usr/share/openshift/linux_$TARGETARCH/oc.rhel8 /usr/bin/oc
+
+RUN ln -s /usr/bin/oc /usr/bin/kubectl
 
 RUN yum install -y httpd-tools
 
--- a/cmd/generate/dockerfile-templates/BuildImageDockerfile.template
+++ b/cmd/generate/dockerfile-templates/BuildImageDockerfile.template
@@ -1,16 +1,13 @@
 # DO NOT EDIT! Generated Dockerfile.
 
+FROM registry.ci.openshift.org/ocp/4.17:tools as tools
+
 # Dockerfile to bootstrap build and test in openshift-ci
 FROM {{.builder}} as builder
 
-RUN echo "[kubernetes]" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "name=Kubernetes" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "baseurl=https://pkgs.k8s.io/core:/stable:/v1.29/rpm/" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "enabled=1" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "gpgcheck=1" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "gpgkey=https://pkgs.k8s.io/core:/stable:/v1.29/rpm/repodata/repomd.xml.key" >> /etc/yum.repos.d/kubernetes.repo
+COPY --from=tools /usr/bin/oc /usr/bin/kubectl /usr/bin/
 
-RUN yum install -y kubectl httpd-tools
+RUN yum install -y httpd-tools
 
 RUN wget https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 && \
     chmod 700 ./get-helm-3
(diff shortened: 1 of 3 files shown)


cmd/generate/BuildImageDockerfile.template
```diff
@@ -1,16 +1,13 @@
 # DO NOT EDIT! Generated Dockerfile.
 
+FROM registry.ci.openshift.org/ocp/4.17:tools as tools
```
mgencur, line 3: I'm wondering if there's a way to not hardcode `4.17` here.

creydr: I didn't see any "latest" (but also don't know a way to browse the registry). We could also try to get it from the builder image name, but that's unsafe I guess.

mgencur: Just an idea. Maybe put another template next to this one? https://github.com/openshift-knative/hack/blob/main/cmd/generate/generate.go#L88
I know we already have a few arguments but having it sid-by-side with the golang builder could ensure that we will be updating it at the same time.

mgencur: But we can also leave it as it is. This will be changing rarely.

creydr: I use a constant now for the image and added a comment with some details & "update recommendations" for it

matejvasek:
```
docker run -it --rm localhost/kubectl-oc:latest oc
oc: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by oc)
```

pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
```diff
@@ -1,16 +1,13 @@
 # DO NOT EDIT! Generated Dockerfile.
```
matejvasek, line 1: I tried this Dockerfile and got:
```
docker run -it --rm localhost/kubectl-oc:latest oc
oc: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by oc)
```

cmd/generate/BuildImageDockerfile.template
matejvasek: **NOTE:** It's important that the tools image and the builder image use same version of glibc.

matejvasek: or at least the builder image has newer glibc than tool image.

pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
matejvasek: Then again 4.15 would not be in sync with cluster version, although it would not cause trouble most likely.

matejvasek: Maybe we could download oc from somewhere else? OCP cluster serves it at some HTTP endpoint?

creydr: 4.15 is based on RHEL8 too.

matejvasek: or download it from https://access.redhat.com

creydr: So seems not the OCP alignment with the builder image is important, but the RHEL version

matejvasek: yep, more specifically glibc version

matejvasek: Still it would be nice to hat ocp version also aligned.

matejvasek: builder.glibc >= tools.glibc must hold

matejvasek: > OCP cluster serves it at some HTTP endpoint?

I am afraid that `oc` would also use too new glibc.

matejvasek: @dsimansk give us a hint:
https://github.com/openshift-knative/kn-plugin-event/blob/main/openshift/ci-operator/build-image/Dockerfile#L8

cmd/generate/dockerfile-templates/BuildImageDockerfile.template
```diff
@@ -1,16 +1,17 @@
...
 
-RUN echo "[kubernetes]" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "name=Kubernetes" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "baseurl=https://pkgs.k8s.io/core:/stable:/v1.29/rpm/" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "enabled=1" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "gpgcheck=1" >> /etc/yum.repos.d/kubernetes.repo && \
-    echo "gpgkey=https://pkgs.k8s.io/core:/stable:/v1.29/rpm/repodata/repomd.xml.key" >> /etc/yum.repos.d/kubernetes.repo
+ARG TARGETARCH
+
+COPY --from=tools /usr/share/openshift/linux_$TARGETARCH/oc.rhel8 /bin/oc
```
matejvasek, line 10: `/bin/oc => /usr/bin/oc` since following lines assume that path

matejvasek:
The test file seems to be not updated.

pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
```diff
@@ -1,16 +1,17 @@
...
 
+FROM registry.ci.openshift.org/ocp/4.17:cli-artifacts as tools
```
matejvasek, line 3: For now this is OK since we support only amd64 but there should be `--platform=linux/amd64` because the registry.ci.openshift.org/ocp/4.17:cli-artifacts is not multiarch.

matejvasek: This comment probably should have been on the template file.

cmd/generate/dockerfile-templates/BuildImageDockerfile.template
creydr: Good catch. Updated

PR #245, opened 2024-09-05
Add possibility to run go mod download

creydr:
Add an option to run `go mod download` in the builder image.

Suggestions on a better flag name (instead of `use-vendor=false`) are welcome :D

### DIFF
--- a/cmd/generate/Dockerfile.template
+++ b/cmd/generate/Dockerfile.template
@@ -4,6 +4,12 @@ ARG GO_RUNTIME=registry.access.redhat.com/ubi8/ubi-minimal
 
 FROM $GO_BUILDER as builder
 
+{{- if eq .use_vendor false }}
+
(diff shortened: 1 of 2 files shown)


cmd/generate/Dockerfile.template
```diff
@@ -4,6 +4,12 @@ ARG GO_RUNTIME=registry.access.redhat.com/ubi8/ubi-minimal
 
 FROM $GO_BUILDER as builder
 
+{{- if eq .use_vendor false }}
+
+COPY go.mod go.sum ./
+RUN go mod download -x
+{{- end }}
+
 COPY . .
```
creydr, line 13: This is probably mostly run in CI. Anyhow to save the devs some time, I added it before the `COPY . .`

creydr:
/hold
As #246 probably helps already with this

PR #248, opened 2024-09-06
Allow to specify template to use for Dockerfile

creydr:
* Add `template-name` option to allow to specify which template to use for the Dockerfile
* Add template for func-util
* Remove `symlink-names` option again, as this was included in the func-util template directly

/hold until we have feedback from @matejvasek if func-util image works in https://redhat-internal.slack.com/archives/CKR568L8G/p1725614736094379?thread_ts=1725374750.726439&cid=CKR568L8G


cmd/generate/dockerfile-templates/FuncUtilDockerfile.template
creydr, line 1: @matejvasek fyi

PR #249, opened 2024-09-09
Fix `spec.componentName` for long names

pierDipi:


pkg/konfluxgen/konfluxgen.go
```diff
@@ -416,8 +416,17 @@ func Sanitize(input interface{}) string {
...
+	)
+
 	in := fmt.Sprintf("%s", input)
-	// TODO very basic name sanitizer
+	if len(in) <= longest {
+		return in
+	}
+
+	in = strings.ReplaceAll(in, "release-v", "")
+	in = strings.ReplaceAll(in, "release-", "")
```
creydr, line 429: maybe we remove this for all (not only, when it's too long (for consistency)) :shrug:

pierDipi: yeah, that forces to delete all existing resources, while we need to delete only the ones with too long names

pierDipi: should I delete it everywhere? we would need to delete all components / applications

creydr: ImageRepositories probably too. As it's only eventing components ATM and we have the autoapply, I'd be fine with that and would prefer consistency.

creydr (approved):
/lgtm

PR #253, opened 2024-09-09
Remove unneeded empty line in output default Dockerfile

creydr:
Only cosmetic, to not have two empty lines in generated Dockerfile. For example like [here](https://github.com/openshift-knative/serverless-operator/blob/d7aa65bbe31a518b8658a24047bceca757f9d249/knative-operator/Dockerfile#L16-L19)


mgencur:
@creydr Can you try if it break the patch here? https://github.com/openshift-knative/serverless-operator/blob/main/knative-operator/dockerfile.patch 
This is again around the same line

PR #254, opened 2024-09-09

serverless-qe:


pierDipi (approved):
/lgtm
/approve

PR #258, opened 2024-09-11
Fix: Push changes from release-next cleanup

creydr:
As per title to address https://github.com/openshift-knative/hack/actions/runs/10806641768/job/29975862579


pierDipi (approved):
/lgtm

PR #259, opened 2024-09-11
Add Client to auto-generated CI config

dsimansk:
Hi folks, I'm looking at adding client to CI config generation. See my other comment inline. 

Generated config PR: https://github.com/openshift/release/pull/56532

Image changes in midstream: https://github.com/openshift-knative/client/commit/8bf49361b76f1f4701bd1d3dc177e916fc5bd779

/cc @pierDipi @mgencur

### DIFF
--- a/config/client.yaml
+++ b/config/client.yaml
@@ -1,0 +1,25 @@
+config:
+  branches:
+    release-v1.15:
+      openShiftVersions:
+        - version: "4.15"
+          skipCron: true
+repositories:
+  - dockerfiles: {}
+    e2e:
+      - match: .*e2e$
+        skipCron: true
+        skipImages:
+            - knative-client-kn
+            - knative-client-cli-artifacts
+    ignoreConfigs:
+      matches:
+        - .*release-next*
+        - .*release-v1.*.yaml$
+    imagePrefix: knative-client
+    org: openshift-knative
+    promotion:
+      namespace: openshift
+    repo: client
+    canonicalGoRepository: github.com/knative/client
+    slackChannel: '#serverless-ci'
--- a/.github/workflows/release-generate-ci.yaml
+++ b/.github/workflows/release-generate-ci.yaml
@@ -30,6 +30,13 @@ jobs:
           repository: openshift-knative/backstage-plugins
           token: ${{ secrets.SERVERLESS_QE_ROBOT }}
       - if: ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule') && github.ref_name == 'main' }}
+        name: '[client] Clone repository'
+        uses: actions/checkout@v4
+        with:
+          path: ./src/github.com/openshift-knative/hack/openshift-knative/client
+          repository: openshift-knative/client
+          token: ${{ secrets.SERVERLESS_QE_ROBOT }}
+      - if: ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule') && github.ref_name == 'main' }}
         name: '[eventing-hyperfoil-benchmark] Clone repository'
         uses: actions/checkout@v4
         with:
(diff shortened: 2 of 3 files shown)


config/client.yaml
```diff
@@ -0,0 +1,56 @@
+config:
+  branches:
+    release-v1.15:
+      openShiftVersions:
+        - version: "4.15"
+          skipCron: true
```
dsimansk, line 6: Is there a way to skip next OCP version. In this case 4.16, we really don't need to retest `kn` on multiple versions in the midstream.

config/client.yaml
```diff
@@ -0,0 +1,56 @@
...
+    ignoreConfigs:
+      matches:
+        - .*release-next*
+        - .*release-v1.*.yaml$
+    customConfigs:
+      - releaseBuildConfiguration:
+          binary_build_commands: |
+            make install
+            make build-cross
+          test_binary_build_commands: make test install
```
dsimansk, line 24: This's supported by the change in `prowgen_config.go` from L263. 

So far I've added a generated Dockerfiles to client `release-v1.15`. We used this build binary mechanism since the forever. See the custom Dockerfile.cliartifacts and the injection. 

I'll look into reworking the image to have its own build phase, but I'm not quiet sure how to wire it with generator as it's additional image definition.

pkg/prowgen/prowgen_config.go
```diff
@@ -53,6 +53,7 @@ type IgnoreConfigs struct {
 
 type Promotion struct {
 	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+	SkipSrc   bool   `json:"skipSrc,omitempty" yaml:"skipSrc,omitempty"`
```
dsimansk, line 56: Added this `bool` to optionally skip src image promotion. We don't utilize it in client.

pkg/prowgen/prowgen_config.go
```diff
@@ -257,6 +258,17 @@ func NewGenerateConfigs(ctx context.Context, r Repository, cc CommonConfig, opts
...
+			// Used in client.yaml config
+			for _, customCfg := range r.CustomConfigs {
+				if len(customCfg.ReleaseBuildConfiguration.BinaryBuildCommands) > 0 {
+					cfg.BinaryBuildCommands = customCfg.ReleaseBuildConfiguration.BinaryBuildCommands
+				}
+				if len(customCfg.ReleaseBuildConfiguration.TestBinaryBuildCommands) > 0 {
+					cfg.TestBinaryBuildCommands = customCfg.ReleaseBuildConfiguration.TestBinaryBuildCommands
+				}
+			}
+
```
dsimansk, line 271: This generate build commands based on the following yaml. It's custom config, but it needs to be in root of yaml doc.

```
- releaseBuildConfiguration:
          binary_build_commands: |
            make install
            make build-cross
          test_binary_build_commands: make test install
```

config/client.yaml
pierDipi: 4.16 is only an "on demand" test and it added automatically

dsimansk: Ahh ok, makes sense then.

config/client.yaml
pierDipi: cc @mgencur

pkg/prowgen/prowgen_config.go
dsimansk: Removed this bit, I'm testing out a new `Dockerfile` for artifacts image, that uses OpenShift CI pipeline src.

dsimansk: https://github.com/openshift-knative/client/commit/3f329251741e3917e28e3b461d9c52d7afec9028#diff-6492e3702be91832f2bf8bb9063894f0849b3a8189b2b3622b41198fda31830a

pierDipi: nice!

dsimansk:
OK, so the build test is passing now. The integration tests are running now. IMO the config is ready to be merged once those are done.

https://github.com/openshift/release/pull/56532

/cc @pierDipi @creydr

dsimansk:
As per our call. The PR is ready to be merged now.

pierDipi (approved):
/lgtm
/approve

PR #260, opened 2024-09-11
Remove chmod on `/etc/passwd` in BuildImageDockerfile

creydr:
Seems not to be needed anymore ([slack](https://redhat-internal.slack.com/archives/CD87JDUB0/p1726045452787339)) and prevents OCP to add the user to /etc/passwd which causes issues on kn-plugin-func tests


pierDipi (approved):
/lgtm

PR #266, opened 2024-09-11
Discover: Use Konflux config from latest release branch

creydr:
As per title

### DIFF
--- a/pkg/discover/discover.go
+++ b/pkg/discover/discover.go
@@ -98,7 +98,12 @@ func discover(ctx context.Context, path string) error {
 			if latestConfigured == availableBranches[i] {
 				for ; i < len(availableBranches); i++ {
 					if _, ok := inConfig.Config.Branches[availableBranches[i]]; !ok {
-						inConfig.Config.Branches[availableBranches[i]] = inConfig.Config.Branches[latest]
+						// use branch config from "latest" branch (can be release-next)
+						branchConfig := inConfig.Config.Branches[latest]
+						// take Konflux config from latestConfigured branch and not from release-next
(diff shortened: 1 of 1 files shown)


pkg/discover/discover.go
```diff
@@ -98,7 +98,12 @@ func discover(ctx context.Context, path string) error {
 			if latestConfigured == availableBranches[i] {
 				for ; i < len(availableBranches); i++ {
 					if _, ok := inConfig.Config.Branches[availableBranches[i]]; !ok {
-						inConfig.Config.Branches[availableBranches[i]] = inConfig.Config.Branches[latest]
+						// use branch config from "latest" branch (can be release-next)
+						branchConfig := inConfig.Config.Branches[latest]
+						// take Konflux config from latestConfigured branch and not from release-next
+						branchConfig.Konflux = inConfig.Config.Branches[latestConfigured].Konflux
+
+						inConfig.Config.Branches[availableBranches[i]] = branchConfig
```
pierDipi, line 106: Since new branches will be 1.16+, and Konflux is *required* can we just do

```
						// use branch config from "latest" branch (can be release-next)
						branchConfig := inConfig.Config.Branches[latest]
						branchConfig.Konflux.Enabled = true
```
and we would keep the release-next Konflux configuration there but `enabled: false`

The reasoning being, we can "maintain" the configuration there without touching "existing" branches

creydr: I thought about this initially too, but was not sure, if everything under config/ will be built on Konflux (e.g. backstage plugins)

creydr: Ok. Updated

pierDipi (approved):
/lgtm
/approve

PR #278, opened 2024-09-13
[generate] Create RPMs lock file for func-utils and must-gather

pierDipi:
This is related to hermetic builds, see the Jira for more details https://issues.redhat.com/browse/SRVCOM-3305

### DIFF
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -81,6 +81,7 @@ func main() {
 		imagesFromRepositoriesURLFmt string
 		additionalPackages           []string
 		templateName                 string
+		rpmsLockFileEnabled          bool
 	)
 
 	defaultIncludes := []string{
@@ -109,6 +110,7 @@ func main() {
 	pflag.StringVar(&imagesFromRepositoriesURLFmt, "images-from-url-format", "https://raw.githubusercontent.com/openshift-knative/%s/%s/openshift/images.yaml", "Additional images to be pulled from other midstream repositories matching the tag in project.yaml")
 	pflag.StringArrayVar(&additionalPackages, "additional-packages", nil, "Additional packages to be installed in the image")
 	pflag.StringVar(&templateName, "template-name", defaultDockerfileTemplateName, fmt.Sprintf("Dockerfile template name to use. Supported values are [%s, %s]", defaultDockerfileTemplateName, funcUtilDockerfileTemplateName))
+	pflag.BoolVar(&rpmsLockFileEnabled, "generate-rpms-lock-file", false, "Enable the creation of the rpms.lock.yaml file")
 	pflag.Parse()
 
 	if rootDir == "" {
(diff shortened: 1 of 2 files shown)


cmd/generate/generate.go
```diff
@@ -103,6 +110,7 @@ func main() {
 	pflag.StringVar(&imagesFromRepositoriesURLFmt, "images-from-url-format", "https://raw.githubusercontent.com/openshift-knative/%s/%s/openshift/images.yaml", "Additional images to be pulled from other midstream repositories matching the tag in project.yaml")
 	pflag.StringArrayVar(&additionalPackages, "additional-packages", nil, "Additional packages to be installed in the image")
 	pflag.StringVar(&templateName, "template-name", defaultDockerfileTemplateName, fmt.Sprintf("Dockerfile template name to use. Supported values are [%s, %s]", defaultDockerfileTemplateName, funcUtilDockerfileTemplateName))
+	pflag.BoolVar(&rpmsLockFileEnabled, "generate-rpms-lock-file", false, "Enable the creation of the rpms.lock.yaml file")
```
creydr, line 113: Is there a reason, why you add this as a parameter? Isn't this only relevant for func? And not sure if we should use "dedicated" lock files for the different components - or would you for simplicity use one big, which is used by multiple components?

creydr (commented):
Thanks for working on this.
Had only 2 questions

cmd/generate/rpms.lock.yaml
```diff
@@ -0,0 +1,61 @@
...
+        size: 330744
+        checksum: "sha256:b03b86ac05bd5871f14e00400a6ddd30d8e9ed97b93e56e96789706c36e2f379"
+      - repoid: ubi-8-baseos-rpms
+        url: "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/x86_64/baseos/os/Packages/t/tar-1.30-9.el8.x86_64.rpm"
+        size: 858812
+        checksum: "sha256:6b0dc7341d743c89fa038292a7e04761ebb6cc98208ebc26dee9f01e2c1a9529"
+      - repoid: ubi-8-baseos-rpms
+        url: "https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/x86_64/baseos/os/Packages/r/rsync-3.1.3-19.el8_7.1.x86_64.rpm"
+        size: 420156
+        checksum: "sha256:7aef9de61fbf590995b07d92f99a3f3478d6c0543d7a6e1ebb6f4b1c02334283"
```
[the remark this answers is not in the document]
pierDipi: yes, this is needed for must gather

cmd/generate/generate.go
pierDipi: for must gather

pierDipi: dedicated lockfile is not worth it, let's optimize when we actually need to, currently, we only have a few RPMs across the board

cmd/generate/rpms.lock.yaml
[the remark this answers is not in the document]
creydr: :+1:

cmd/generate/generate.go
creydr: :+1:

PR #280, opened 2024-09-13
[konfluxgen] Enable RPMs and Go pre-fetching for hermetic builds

pierDipi:
This in combination with https://github.com/openshift-knative/hack/pull/278, will enable func-utils and must-gather hermetic builds

### DIFF
--- a/pkg/prowgen/prowgen_konflux.go
+++ b/pkg/prowgen/prowgen_konflux.go
@@ -2,7 +2,9 @@ package prowgen
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/fs"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


pkg/prowgen/prowgen_konflux.go
```diff
@@ -96,6 +96,19 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
 						log.Printf("[%s] created nudges (%v) - operatorVersions: %#v - downstreamVersion: %v): %#v", r.RepositoryDirectory(), ok, operatorVersions, downstreamVersion, nudges)
 					}
 
+					prefetchDeps := konfluxgen.PrefetchDeps{}
+					if _, err := os.Stat(filepath.Join(r.RepositoryDirectory(), "rpms.lock.yaml")); err == nil {
+						// If rpms.lock.yaml is present enable dev-package-managers and RPM caching
+						prefetchDeps.DevPackageManagers = "true"
+						prefetchDeps.WithRPMs()
+					}
+					if _, err := os.Stat(filepath.Join(r.RepositoryDirectory(), "vendor")); err != nil {
```
creydr, line 105: not sure, if os.Stat can return multiple errors, but maybe we can check if it is "FileNotExists" error (`os.IsNotExist(err)`)

pierDipi: Done

pierDipi: using the suggested improvement
> // This function predates errors.Is. It only supports errors returned by
// the os package. New code should use errors.Is(err, fs.ErrNotExist).

creydr (approved):
/lgtm

PR #282, opened 2024-09-16
Use `params.hermetic` to configure hermetic in FBC builder pipeline

pierDipi:
Related to https://github.com/konflux-ci/build-definitions/issues/1426


creydr (approved):
/lgtm

PR #287, opened 2024-09-18
Add kn-plugin-func repo

creydr:
Adding the kn-plugin-func repo, so we have:

* automatic prow config generation (changes can be seen for example [here](https://github.com/openshift/release/compare/master...creydr:openshift-release:sync-serverless-ci-func?expand=1))
* possibility to generate Konflux manifests

**Hint:** 
* kn-plugin-func is one version ahead the other repos. This means for example client 1.15 includes func 1.16. Therefor the "exception" was added in e3942d0c2faf724cb1ed3e6661f9ff0e9ab45611
* As kn-plugin-func does not yet have a 1.16 branch, in which Konflux would be enabled, no branch has yet Konflux enabled for func (as we have the automatic upstream branch mirroring in place for func [now](https://github.com/openshift-knative/kn-plugin-func/pull/890), the next "discover branches" run will detect the 1.16 branch and enable Konflux (as this is enabled for all new branches [now](https://github.com/openshift-knative/hack/pull/266)))
* as the CI tests (unit & e2e-oncluster-tests) are currently configured directly in prow and not only via Makefile targets, the tests are added directly in `functions.yaml` (`.repositories[].tests[]`), but I guess this can be changed at some point (@jrangelramos fyi)


config/functions.yaml
pierDipi, line 1: For consistency, should we call this `kn-plugin-func` ?

creydr: :+1: 
updated

pierDipi (approved):
/lgtm
/approve

PR #288, opened 2024-09-18

serverless-qe:

### DIFF
--- a/config/kn-plugin-func.yaml
+++ b/config/kn-plugin-func.yaml
@@ -2,12 +2,12 @@ config:
   branches:
     release-next:
       openShiftVersions:
-      - version: "4.15"
-        skipCron: true
+      - skipCron: true
+        version: "4.15"
     serverless-1.34:
       openShiftVersions:
-      - version: "4.15"
-        skipCron: true
+      - skipCron: true
+        version: "4.15"
 repositories:
 - dockerfiles: {}
   ignoreConfigs: {}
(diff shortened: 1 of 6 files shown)


config/kn-plugin-func.yaml
```diff
@@ -1,13 +1,145 @@
 config:
   branches:
+    "":
+      konflux:
+        enabled: true
+      openShiftVersions:
+      - skipCron: true
+        version: "4.15"
```
creydr, line 8: ?

creydr: Addressed in https://github.com/openshift-knative/hack/pull/289

config/kn-plugin-func.yaml
creydr, line 1: @pierDipi I tend to revert https://github.com/openshift-knative/hack/pull/287 and recreate a PR and set in on `/hold` until we have release-1.16 as branch configured, otherwise pipeline updates for other repos is blocked (branch discovery should though happen for the same time for all repos, but we don't need < 1.16 in func).
We could also check to add some exceptions in code for func for this, but I think that's not worth and only temporary IMO.
WDYT?

pierDipi: why not trying to fix it?

pierDipi: I think it's an issue with branch naming and comparing them as we didn't expect `serverless-*` ?

https://github.com/openshift-knative/hack/blob/e6e2f70ccaddecc6f803f12e51b2af88f129e570/pkg/prowgen/prowgen_git.go#L58-L88

pierDipi: but I'm ok either way

creydr: https://github.com/openshift-knative/hack/pull/291 (my only concern was to add this code only because of the func serverless-1.34 branch)

config/eventing-istio.yaml
```diff
@@ -29,6 +29,7 @@ config:
     release-v1.15:
       konflux:
         enabled: true
+        imageOverrides: null
```
creydr, line 32: https://github.com/openshift-knative/hack/pull/292 should help

config/serverless-operator.yaml
```diff
@@ -13,14 +13,13 @@ config:
       - onDemand: true
         version: "4.13"
     release-1.35:
-      prowgen:
-        enabled: false
```
creydr, line 17: @mgencur I guess, this is not intended to be removed? Probably because Go default value for boolean is false and thus "can be removed". Maybe rename it to `disabled: true`?

mgencur: No. It should stay there. No idea why this is being removed at this point.

creydr: Probably as "false" is the default for Golangs boolean values

creydr: That's what I meant with renaming it to "disabled" and requiring a "true" value

mgencur: This probably needs some additional work here: https://github.com/openshift-knative/hack/blob/main/pkg/discover/discover.go#L97
Looking into it.

mgencur: I think I will just rename it as you suggested.

mgencur: Sent in https://github.com/openshift-knative/hack/pull/294

creydr (approved):
/lgtm

PR #290, opened 2024-09-18
Define image overrides in serverless-operator configuration

mgencur:
Fixes https://issues.redhat.com/browse/SRVCOM-3350

### DIFF
--- a/config/serverless-operator.yaml
+++ b/config/serverless-operator.yaml
@@ -14,12 +14,12 @@ config:
         version: "4.13"
     release-1.35:
       konflux:
-        enabled: true
-      imageOverrides:
-        - name: GO_BUILDER
-          pullSpec: "brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_golang_1.22"
-        - name: GO_RUNTIME
-          pullSpec: "registry.access.redhat.com/ubi8/ubi-minimal"
+        enabled: false
+        imageOverrides:
+          - name: GO_BUILDER
+            pullSpec: "brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_golang_1.22"
+          - name: GO_RUNTIME
+            pullSpec: "registry.access.redhat.com/ubi8/ubi-minimal"
 repositories:
 - customConfigs:
   - name: aws-ovn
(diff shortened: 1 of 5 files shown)


config/serverless-operator.yaml
```diff
@@ -12,6 +12,16 @@ config:
         version: "4.15"
       - onDemand: true
         version: "4.13"
+    release-1.35:
+      prowgen:
+        enabled: false
+      konflux:
+        enabled: false
```
mgencur, line 19: After the branch is actually created we'd flip this single flag to true. But we'd still keep prowgen.enabled == false

config/serverless-operator.yaml
```diff
@@ -12,6 +12,16 @@ config:
...
+        imageOverrides:
```
mgencur, line 20: It is here so that Konflux generator for other repositories (Eventing, etc.) can read it from here.

config/serverless-operator.yaml
pierDipi: SO is special cased for the flag https://github.com/openshift-knative/hack/blob/d565e2e9fbe0c39f0570ca737ebd3c0afc774683/pkg/prowgen/prowgen_konflux.go#L169-L199

so we don't even need this flag configuration nor flip it

pierDipi (commented):
Looks good, just one comment / suggestion

pkg/prowgen/prowgen_konflux.go
```diff
@@ -83,11 +83,21 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
...
 					}
 					log.Println("Version label:", versionLabel)
+
+					soConfig, loadErr := LoadConfig("config/serverless-operator.yaml")
+					if loadErr != nil {
+						return fmt.Errorf("failed to load config for serverless-operator: %w", loadErr)
+					}
+					for soConfigBranchName, br := range soConfig.Config.Branches {
+						if soBranchName == soConfigBranchName {
+							for _, img := range br.Konflux.ImageOverrides {
```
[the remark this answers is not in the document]
mgencur: Yeah. Thanks for suggestions. Pushed.

creydr (commented):
/lgtm

PR #291, opened 2024-09-18
Discover: Respect serverless-x.y.z branch names in sorting too

creydr:
Follow up on https://github.com/openshift-knative/hack/pull/288#discussion_r1764905222 to take funcs current branch naming scheme into account too to discover branches

### DIFF
--- a/pkg/prowgen/prowgen_git.go
+++ b/pkg/prowgen/prowgen_git.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/prowgen/prowgen_git.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift-knative/hack/pkg/soversion"
```
pierDipi, line 7: can we sort imports here?

creydr: updated

PR #296, opened 2024-09-20
Bump OCP 4.15 to 4.16 in configs

mgencur:
https://issues.redhat.com/browse/SRVCOM-3351

Also, make it explicit in the config which version uses cluster pool. Having to update the constant in prowgen_tests.go every time would be annoying, and error prone (we could forget about it).

/hold

Depends on https://github.com/openshift/release/pull/56966


pierDipi (approved):
/lgtm

PR #298, opened 2024-09-24
Don't force push when there are no changes

pierDipi:
Prevent force pushing when there are no changes to avoid
unnecessary CI runs.

First commit is the logic change, second generates the action

### DIFF
--- a/.github/workflows/release-generate-ci-template.yaml
+++ b/.github/workflows/release-generate-ci-template.yaml
@@ -85,8 +85,20 @@ jobs:
         run: |
           set -x
           git remote add fork "https://github.com/serverless-qe/release.git"
-          git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/release.git" sync-serverless-ci:sync-serverless-ci -f
-          gh pr create --base master --head serverless-qe:sync-serverless-ci --fill-verbose || true
+          branch="sync-serverless-ci"
+          remote_exists=$(git ls-remote --heads fork "$branch")
+          if [ -z "$remote_exists" ]; then
+            # remote doesn't exist.
+            git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/release.git" "$branch:$branch" -f
+            exit $?
+          fi
+          git fetch fork "$branch"
+          if git diff --quiet "fork/$branch" "$branch"; then
+            echo "Branches are identical. No need to force push."
+          else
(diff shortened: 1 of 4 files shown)


pkg/prowgen/prowgen_konflux.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	KonfluxBranchPrefix = "sync-konflux/"
+	KonfluxBranchPrefix = "sync-konflux-"
```
pierDipi, line 24: This change is because in Git we cannot have both branch 'foo' and branch 'foo/bar'

.github/workflows/release-generate-ci-template.yaml
```diff
@@ -85,8 +85,20 @@ jobs:
...
-          git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/release.git" sync-serverless-ci:sync-serverless-ci -f
-          gh pr create --base master --head serverless-qe:sync-serverless-ci --fill-verbose || true
+          branch="sync-serverless-ci"
+          remote_exists=$(git ls-remote --heads fork "$branch")
+          if [ -z "$remote_exists" ]; then
+            # remote doesn't exist.
+            git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/release.git" "$branch:$branch" -f
+            exit $?
+          fi
+          git fetch fork "$branch"
```
mgencur, line 95: Should this be in the `else` branch? Not sure what happens if the remote branch doesn't exist. We probalby suppose it always exists?

pkg/action/update_action.go
```diff
@@ -134,17 +134,28 @@ func updateAction(inConfig *prowgen.Config) ([]interface{}, []interface{}) {
...
+  git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/$repo.git" "$branch:$branch" -f
+  exit $?
+fi
+git fetch fork "$branch"
+if git diff --quiet "fork/$branch" "$branch"; then
+  echo "Branches are identical. No need to force push."
+else
+  git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/$repo.git" "$branch:$branch" -f
+fi
+git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/$repo.git" "$branch:$branch" -f
```
mgencur, line 153: This push shouldn't be here, right?

mgencur: Otherwise this PR looks good.

pierDipi: Good catch!

PR #305, opened 2024-09-30
Add workflow in each repo to apply Konflux manifests on changes

creydr:
* add a reusable action to apply the Konflux manifests
* Adds a workflow to each repo & branch which have Konflux enabled calling the above reusable action. This allows to apply the manifests directly after changes in the `.konflux/` directory in the repositories and does not need the periodic runs from hack repo anymore

/hold
as we need to move the `KONFLUX_SA_TOKEN` secret from the hack repo to the org (when we agreed on this PR)

### DIFF
--- a/.github/workflows/apply-konflux-manifests.yaml
+++ b/.github/workflows/apply-konflux-manifests.yaml
@@ -1,6 +1,13 @@
-name: Reusable Apply Konflux Manifests
+name: Apply Konflux Manifests
 on:
-  workflow_call:
+  schedule:
+    - cron: "0 6 * * *" # Daily at 06:00.
+  workflow_dispatch: # Manual workflow trigger
+  push:
+    paths:
+      - ".konflux/**"
+    branches:
+      - 'main'
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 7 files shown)


pierDipi:
why?

pierDipi:
Similar to what we already do for SO, I think we discuss that as the approach we would have taken https://github.com/openshift-knative/serverless-operator/commit/45eccfe5e170b06adb4b85c40d303752d44aba55

creydr:
> why not calling the existing workflow in this repo instead of duplicating it everywhere? That would solve also the possibility for CI config to be different when merging PRs so from each repo we would call the generate CI and Konflux apply

Not sure if I understand you correctly, but isn't this what I am proposing in my PR? I made the existing workflow to apply the konflux manifests "reusable" and only add a new workflow to the other repos, which calls this workflow:

https://github.com/openshift-knative/hack/blob/6cc93a788c9809ba150449eb496a7aeab5d2bb88/pkg/konfluxgen/apply-konflux-manifests-workflow.yaml#L12

This allows us to still manage the core workflow in this repo and run the workflow only when it's needed (on pushes to the downstream repos)

PR #309, opened 2024-10-02
Tag images with exact SO version

pierDipi:
This will help with the release as we can specify the exact patch-level version tag.

As soon as SO bumps the patch version, we would get the dependent components to tag newly built images with the new patch version


creydr (approved):
/lgtm

PR #310, opened 2024-10-02
Remove version skip for kn-plugin-func as versions are aligned now

creydr:
As func versions are now aligned with the other components, we can remove the version skip for func

Discussion: https://redhat-internal.slack.com/archives/CLMP7R2G2/p1727789129599039?thread_ts=1727771460.710709&cid=CLMP7R2G2


pierDipi (approved):
/lgtm

PR #311, opened 2024-10-02
Fix discovery of func release-v1.x branches

creydr:
Builds on top of #310

Currently func also has branches in the pattern `serverless-xyz`. Those branches are not discovered/listed (should be fine, as we have serverless-1.34 already in the config and is the last branch of this pattern before switching to `release-v1.x` pattern).
This leads to issues when we want to discover the new `release-v1.x` branches as `availableBranches` does not contain the `serverless-xyz` branches (while `latestConfigured` is `serverless-xyz`) and then later does not add the branch correctly.

This PR addresses it, by manually adding a `release-y1.x` config for func, so `latestConfigured` will be a `release-v1.x` and thus later newer `release-v1.x` branches added correctly.
Alternatively we could also add the `serverless-*` branches to the list in prowgen.Branches() (`branchesBytes, err := Run(ctx, r, "git", "--no-pager", "branch", "--list", "release-*", "--list", "serverless-*")`), but as this is only specific to func and a one time effort (as `serverless-xyz` branches are deprecated), I thought this has less impact.


pierDipi:
Tests are failing

creydr:
> Tests are failing

:facepalm: 
fixed

pierDipi (approved):
/lgtm
/approve

PR #312, opened 2024-10-02
[discover] Fix SO branch discovery and remove update action SO special casing

pierDipi:
- Fix SO branch discovery by adding "future branches" when the non-existing configured one is created/available
- Remove Update action specifial casing for SO and skip non existing branches
- Run `make discover-branches`
- Add image overrides for SO main

This will also enable more special casing removal in the future.


pierDipi:
@creydr @mgencur can you take a look?

creydr (commented):
/lgtm

PR #314, opened 2024-10-02
Use ose operator registry as index base image

creydr:
Follow up on https://github.com/openshift-knative/serverless-operator/pull/2911#pullrequestreview-2342792014 to use `registry.redhat.io/openshift4/ose-operator-registry` as index base image and pass it as a build arg to the FBC build.

Also needed according to https://gitlab.cee.redhat.com/konflux/docs/users/-/blob/main/topics/releasing/preparing-for-release.adoc#user-content-publishing-a-fbc-graph


pkg/project/serverlessoperator/project.go
```diff
@@ -0,0 +1,58 @@
...
+
+type Kube struct {
+	MinVersion string `yaml:"minVersion"`
+}
+
+type OcpVersion struct {
+	Min   string `yaml:"min"`
+	Max   string `yaml:"max"`
+	Label string `yaml:"label"`
+}
```
pierDipi, line 49: can't we reuse the Metadata struct we already have?

Also I don't think we need all these fields?

creydr: I thought the existing Metadata struct was for a different purpose / structure of project.yaml. But if you're OK to reuse it for this, I am fine too. Will update it...

pkg/prowgen/prowgen_konflux.go
```diff
@@ -260,7 +260,11 @@ func GenerateKonfluxServerlessOperator(ctx context.Context, openshiftRelease Rep
 		if err != nil {
 			return err
 		}
-		buildArgs := []string{fmt.Sprintf("VERSION=%s", soMetadata.Project.Version)}
+
+		buildArgs := []string{
+			fmt.Sprintf("VERSION=%s", soMetadata.Project.Version),
+			fmt.Sprintf("OPM_IMAGE=registry.redhat.io/openshift4/ose-operator-registry-rhel9:v%s", soMetadata.Requirements.OcpVersion.Max),
```
pierDipi, line 266: don't we want to stay on RHEL8 for everything?

creydr: ose-operator-registry has only the tags up to 4.14 (https://catalog.redhat.com/software/containers/openshift4/ose-operator-registry/5cddd0bed70cc57c44b2e1f3), `ose-operator-registry-rhel9` also has the ones afterwards.

mgencur: I have just run across this doc:https://gitlab.cee.redhat.com/konflux/docs/users/-/blob/main/topics/releasing/preparing-for-release.adoc#user-content-publishing-a-fbc-graph
Especially:

> While all of the following RPA samples specify a target OpenShift version with a tag, the actual version (and tag) [are determined by](https://konflux-ci.dev/architecture/ADR/0026-specifying-ocp-targets-for-fbc.html) the parent image used for the FBC build. In order to ship to the proper OCP version, ensure that you are using the properly tagged registry.redhat.io/openshift4/ose-operator-registry parent image

In my opinion, we will need to build multiple different Index images, one for each supported OpenShift version. So, right now it would be 4.13 - 4.17, each version between `min: XX` and `max: YY`. 
I'm saying it here because we probably need to iterate, rather than using just `max`

pierDipi:
See https://github.com/openshift-knative/hack/pull/334

PR #316, opened 2024-10-03
support multi-arch build SRVCOM-3256

rudyredhat1:
Hi all,

I have tried to add the docker-build-oci-ta.yaml file to konfluxgen to support multi-arch.
I am confused with the Generate() and have added the comment on Line 124, i need some help over there.
Rest please have a look and let me where to change or update the values.
Jira ticket = https://issues.redhat.com/browse/SRVCOM-3256

/cc @pierDipi @creydr @mgencur 
Thanks


pierDipi (commented):
do we need to move to TA pipelines? Can't we add the additional steps to the existing docker-build and docker-java-build pipelines ?

PR #320, opened 2024-10-08
Fix SO CEL expression and migrate index Dockerfile

pierDipi:
Related to https://github.com/openshift-knative/serverless-operator/pull/2924


creydr (approved):
/lgtm

PR #325, opened 2024-10-09
Fix konflux-apply to skip non existing branches

creydr:
Fixing konflux-apply to skip non existing branches (https://github.com/openshift-knative/hack/actions/runs/11249266890)

Similar to #312 but for konflux-apply


pierDipi (approved):
/lgtm

PR #326, opened 2024-10-09
Remove patch to remove RPM scans again

creydr:
As per title


pierDipi (commented):
Remove the file `patch_rpms_scan_remove.patch.yaml` too ?

creydr:
> Remove the file `patch_rpms_scan_remove.patch.yaml` too ?

Forgot to add to commit
Updated...

PR #327, opened 2024-10-09
[main] Add Konflux configurations

serverless-qe:
Add Konflux components and pipelines

PR #328, opened 2024-10-09
Fix ReleasePlanAdmission directory name

creydr:
See for example https://github.com/openshift-knative/hack/pull/327#issuecomment-2401839410


pierDipi (approved):
/lgtm
/approve

I guess we could use constants

PR #329, opened 2024-10-09
Fix konflux-apply path

creydr:
`prowgen.Run(ctx, r, ...)` has as working directory r.RepositoryDirectory(). Therefor we need to pass the relative path to r.RepositoryDirectory()


pierDipi (approved):
/lgtm

PR #334, opened 2024-10-11
Add FBC applications & components

creydr:
Adding the applications and components for the FBCs

### DIFF
--- a/pkg/konfluxgen/konfluxgen.go
+++ b/pkg/konfluxgen/konfluxgen.go
@@ -97,7 +97,6 @@ type FBCConfig struct {
 	Metadata cioperatorapi.Metadata
 
 	AdditionalTektonCELExpressions string
-	IsHermetic                     bool
 
 	DockerfilePath string
 	ContextDirPath string
@@ -510,7 +509,7 @@ func GenerateFBCApp(cfg FBCConfig) error {
 		Pipeline:                      FBCBuild,
 		Tags:                          append(cfg.Tags, "latest"),
 		BuildArgs:                     cfg.BuildArgs,
-		Hermetic:                      fmt.Sprintf("%t", cfg.IsHermetic),
+		Hermetic:                      "false",
 		AdditionalTektonCELExpression: cfg.AdditionalTektonCELExpressions,
 	}
 
--- a/pkg/konfluxgen/konfluxgen.go
+++ b/pkg/konfluxgen/konfluxgen.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"strings"
 	"text/template"
@@ -70,12 +71,14 @@ type Config struct {
 
 	ResourcesOutputPathSkipRemove bool
 	ResourcesOutputPath           string
+
+	PipelinesOutputPathSkipRemove bool
 	PipelinesOutputPath           string
 
 	AdditionalTektonCELExpressionFunc func(cfg cioperatorapi.ReleaseBuildConfiguration, ib cioperatorapi.ProjectDirectoryImageBuildStepConfiguration) string
+	NudgesFunc                        func(cfg cioperatorapi.ReleaseBuildConfiguration, ib cioperatorapi.ProjectDirectoryImageBuildStepConfiguration) []string
 
-	NudgesFunc func(cfg cioperatorapi.ReleaseBuildConfiguration, ib cioperatorapi.ProjectDirectoryImageBuildStepConfiguration) []string
-	Nudges     []string
+	Nudges []string
 
 	Tags []string
 
@@ -83,25 +86,8 @@ type Config struct {
 
 	IsHermetic func(cfg cioperatorapi.ReleaseBuildConfiguration, ib cioperatorapi.ProjectDirectoryImageBuildStepConfiguration) bool
 
-	ClusterServiceVersionPath string
-}
-
-type FBCConfig struct {
-	ApplicationName string
-	BuildArgs       []string
-
-	ResourcesOutputPath string
-	PipelinesOutputPath string
-	OCPVersion          string
-
-	Metadata cioperatorapi.Metadata
-
-	AdditionalTektonCELExpressions string
-
-	DockerfilePath string
-	ContextDirPath string
-
-	Tags []string
+	ClusterServiceVersionPath  string
+	AdditionalComponentConfigs []TemplateConfig
 }
 
 type PrefetchDeps struct {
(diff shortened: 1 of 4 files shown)


pkg/prowgen/prowgen_konflux.go
```diff
@@ -341,6 +345,44 @@ func GenerateKonfluxServerlessOperator(ctx context.Context, openshiftRelease Rep
...
 
+func generateFBCApplications(soMetadata *project.Metadata, openshiftRelease Repository, r Repository, branch string, release string, resourceOutputPath string, buildArgs []string) error {
+	soMetadata.Requirements = &project.Requirements{
+		OcpVersion: project.OcpVersion{
+			List: []string{
+				"4.13",
+				"4.17",
+			},
+		},
+	}
```
creydr, line 356: This is only temporary for testing until https://github.com/openshift-knative/serverless-operator/pull/2926/ is in

pkg/konfluxgen/konfluxgen.go
```diff
@@ -428,6 +447,149 @@ func Generate(cfg Config) error {
...
+	}
+
+	// Generate application yaml
+	buf := &bytes.Buffer{}
+	if err := applicationTemplate.Execute(buf, config); err != nil {
+		return fmt.Errorf("failed to execute template for application %q: %w", appKey, err)
+	}
+	if err := os.WriteFile(appPath, buf.Bytes(), 0777); err != nil {
+		return fmt.Errorf("failed to write application file %q: %w", appPath, err)
+	}
```
creydr, line 524: Duplications from [konfluxgen.Generate()](https://github.com/openshift-knative/hack/blob/f2a9e86ff33b53185aabec1a55f84660f3523a15/pkg/konfluxgen/konfluxgen.go#L326). But before refactoring it, I want to align on the general approach first

pkg/konfluxgen/konfluxgen.go
```diff
@@ -428,6 +447,149 @@ func Generate(cfg Config) error {
...
+		ProjectDirectoryImageBuildStepConfiguration: cioperatorapi.ProjectDirectoryImageBuildStepConfiguration{
+			To: cioperatorapi.PipelineImageStreamTagReference(componentKey),
+			ProjectDirectoryImageBuildInputs: cioperatorapi.ProjectDirectoryImageBuildInputs{
+				DockerfilePath: cfg.DockerfilePath,
+				ContextDir:     cfg.ContextDirPath,
+			},
+		},
+		Pipeline:                      FBCBuild,
+		Tags:                          append(cfg.Tags, "latest"),
+		BuildArgs:                     cfg.BuildArgs,
```
creydr, line 512: need to add the OLM_IMAGE arg override too

pkg/konfluxgen/konfluxgen.go
pierDipi: Agree there is a lot of duplication

pierDipi: we could "augment" `collectConfigurations` with an optional function for the additional images/components

pkg/konfluxgen/konfluxgen.go
creydr: done

pkg/konfluxgen/konfluxgen.go
creydr: :+1: 
Updated

pkg/prowgen/prowgen_konflux.go
```diff
@@ -342,10 +347,97 @@ func GenerateKonfluxServerlessOperator(ctx context.Context, openshiftRelease Rep
...
+						Metadata: cioperatorapi.Metadata{
+							Org:    r.Org,
+							Repo:   r.Repo,
+							Branch: branch,
+						},
+						Images: []cioperatorapi.ProjectDirectoryImageBuildStepConfiguration{
+							{
+								To: cioperatorapi.PipelineImageStreamTagReference(fmt.Sprintf("serverless-index-%s-fbc-%s", release, v)),
+								ProjectDirectoryImageBuildInputs: cioperatorapi.ProjectDirectoryImageBuildInputs{
+									DockerfilePath: "Dockerfile",
```
creydr, line 383: This results for example in the following:
```
apiVersion: appstudio.redhat.com/v1alpha1
kind: Component
metadata:
  annotations:
    build.appstudio.openshift.io/pipeline: '{"name":"docker-build","bundle":"latest"}'
  name: serverless-index-135-fbc-413
spec:
  componentName: serverless-index-135-fbc-413
  application: serverless-operator-135-fbc-413
  source:
    git:
      url: https://github.com/openshift-knative/serverless-operator.git
      context: ./olm-catalog/serverless-operator-index/v4.13
      dockerfileUrl: Dockerfile
      revision: main
```

From the DockerfilePath doc:

> DockerfilePath is the path to a Dockerfile in the project to run **relative to the context_dir**.

pkg/prowgen/prowgen_konflux.go
```diff
@@ -342,10 +347,97 @@ func GenerateKonfluxServerlessOperator(ctx context.Context, openshiftRelease Rep
...
-func serverlessIndexNudge(downstreamVersion string) string {
-	return konfluxgen.Truncate(konfluxgen.Sanitize(fmt.Sprintf("%s-%s", "serverless-index", downstreamVersion)))
+func serverlessIndexNudges(downstreamVersion string, ocpVersions []string) []string {
+	indexes := []string{
+		konfluxgen.Truncate(konfluxgen.Sanitize(fmt.Sprintf("%s-%s", "serverless-index", downstreamVersion))),
+	}
+
+	for _, v := range ocpVersions {
+		indexes = append(indexes, konfluxgen.Truncate(konfluxgen.Sanitize(fmt.Sprintf("serverless-index-%s-fbc-%s-index", downstreamVersion, v))))
+	}
```
pierDipi, line 440: This is just the component name, do you know if it is supposed to work when the application is different?

pierDipi: it seems to be working https://github.com/openshift/hac-dev/pull/909

pierDipi (approved):
/lgtm
/approve

pkg/prowgen/prowgen_konflux.go
creydr: I only saw, that the nudge-ref is only a list of strings (without an application ref). As they are in the same namespace, I'd assume so, but not sure.
Let me confirm with them

creydr: > I only saw, that the nudge-ref is only a list of strings (without an application ref). As they are in the same namespace, I'd assume so, but not sure. Let me confirm with them

Ah didn't see your message https://github.com/openshift-knative/hack/pull/334#discussion_r1799206591

pierDipi: from one test I saw
```
 it('should return all components in a namespace', () => {
    useK8sWatchResourceMock.mockReturnValue([mockComponentsData, true, undefined]);
    const { result } = renderHook(() => useAllComponents('test-ns'));
    const [components] = result.current;
    expect(components).toHaveLength(3);
  });
```
so it should work

mgencur:
Should we also remove this line? https://github.com/openshift-knative/hack/pull/334/files#diff-888ae255f82f46f198de233d2aa02b58b2f97fcc52b07f89ea8cac5d79514ca4R298
The point is to stop generating the Component and Tekton pipeline for FBC in the `serverless-operator-135` application (note the missing OCP version) and only generate it in the OCP specific applications.

pierDipi:
@mgencur I think more that removing that one, we want to add an exclude otherwise will be anyway added as normal component (since it's part of the CI config)

creydr:
https://github.com/openshift-knative/hack/pull/342

PR #336, opened 2024-10-14
Fix Pipelineruntemplate

creydr:
As per title.

See e.g. https://github.com/openshift-knative/backstage-plugins/pull/60/files#r1799046257

/cc @pierDipi


pierDipi (approved):
/lgtm
/approve

PR #337, opened 2024-10-14
Place CEL expression on a single line to avoid YQ changes

pierDipi:
I'm trying to include the SO operator patch release flow in Konflux builds and I need to use YQ on pipelines to change the tag where the images are pushed


creydr (approved):
/lgtm

PR #340, opened 2024-10-14
Per application RPA

pierDipi:
the RPA has the filename that is not application name dependent, so new applications will override the existing ones rather than creating new ones, so that we can just take them as they are and commit to "`releng`".

Remove `v` prefix from tags.

### DIFF
--- a/.konflux/releaseplanadmissions/serverless-operator-135-1350-prod.yaml
+++ b/.konflux/releaseplanadmissions/serverless-operator-135-1350-prod.yaml
@@ -1,7 +1,7 @@
 apiVersion: appstudio.redhat.com/v1alpha1
 kind: ReleasePlanAdmission
 metadata:
-  name: serverless-operator-135
+  name: serverless-operator-135-1350
   namespace: rhtap-releng-tenant
 spec:
   applications: [serverless-operator-135]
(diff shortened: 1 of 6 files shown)


creydr (commented):
Thanks for working on this. We probably need to adjust the RPAs metadata.name to, to include the patch version (https://redhat-internal.slack.com/archives/CKR568L8G/p1728642911235589?thread_ts=1728642740.140649&cid=CKR568L8G)

pierDipi:
@creydr Done

.konflux/releaseplanadmissions/serverless-operator-135-1350-prod.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: appstudio.redhat.com/v1alpha1
 kind: ReleasePlanAdmission
 metadata:
-  name: serverless-operator-135
+  name: serverless-operator-135-1350
```
creydr, line 4: /lgtm

I am fine with both, but IMO `serverless-operator-1350` would be clearer here. But we can also do this as a follow up
/hold in case you want to address it

pierDipi: I think it's fine?

pierDipi: better than what we have now, so we can improve later

/unhold

pierDipi: changing that is not that simple as the app name includes the major/minor version

PR #341, opened 2024-10-14
Use correct pipeline for OCP FBC builds

creydr:
Fix for https://github.com/openshift-knative/serverless-operator/pull/2937/files#r1799341897


pkg/prowgen/prowgen_konflux.go
```diff
@@ -391,6 +391,9 @@ func generateFBCApplications(soMetadata *project.Metadata, openshiftRelease Repo
 			ComponentNameFunc: func(cfg cioperatorapi.ReleaseBuildConfiguration, ib cioperatorapi.ProjectDirectoryImageBuildStepConfiguration) string {
 				return string(ib.To)
 			},
+			FBCImages: []string{
+				fmt.Sprintf("serverless-index-%s-fbc-%s", release, v),
+			},
```
pierDipi, line 396: why not using the config file ?

creydr: The name for these indexes come only from us (not like for example from the CI config)

pierDipi (approved):
/lgtm

PR #342, opened 2024-10-14
Only build index for OCP versioned apps

creydr:
Follow up on https://github.com/openshift-knative/hack/pull/334#issuecomment-2410846444

### DIFF
--- a/pkg/prowgen/prowgen_konflux.go
+++ b/pkg/prowgen/prowgen_konflux.go
@@ -144,7 +144,6 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
 						},
 						Excludes:            b.Konflux.Excludes,
 						ExcludesImages:      b.Konflux.ExcludesImages,
-						FBCImages:           b.Konflux.FBCImages,
 						JavaImages:          b.Konflux.JavaImages,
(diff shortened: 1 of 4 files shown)


pkg/prowgen/prowgen_konflux.go
```diff
@@ -295,7 +288,6 @@ func GenerateKonfluxServerlessOperator(ctx context.Context, openshiftRelease Rep
 			},
 			Excludes:       b.Konflux.Excludes,
 			ExcludesImages: b.Konflux.ExcludesImages,
-			FBCImages:      b.Konflux.FBCImages,
```
pierDipi, line 298: I would leave it here

pierDipi: in https://github.com/openshift-knative/hack/pull/341, it looks like we need it

creydr: I removed it only from the [`prowgen.Konflux`](https://github.com/openshift-knative/hack/blob/295c5f17bcbc79d3d0035737f6f4abb467bc70b8/pkg/prowgen/prowgen_config.go#L81) struct, but kept it in the [`konfluxgen.Config`](https://github.com/openshift-knative/hack/blob/295c5f17bcbc79d3d0035737f6f4abb467bc70b8/pkg/konfluxgen/konfluxgen.go#L59) struct.

PR #343, opened 2024-10-14
Add serverless-index to default excludeImages of SO

creydr:
Add serverless-index to default excludes, to not override default list (which includes for example the `source` images)


pierDipi (approved):
/lgtm
/approve

PR #350, opened 2024-10-15
Use path-context param in pipeline when specified

creydr:
In case the build context is not the repo root, we need to set it in the pipeline via the `patch-context` (e.g. for the FBC components).

Addresses https://github.com/openshift-knative/serverless-operator/pull/2937#pullrequestreview-2369523887


pierDipi (approved):
/lgtm
/approve

PR #351, opened 2024-10-15
Remove workspace pvc for container builds except FBC builds

pierDipi:
Resolves the PVC errors.

PVC is not needed for TA pipelines


creydr (commented):
/lgtm

PR #352, opened 2024-10-15
Only build images for x86 on PRs

pierDipi:
Multi-platform builds are slow and on PRs don't make a lot of sense.


creydr (approved):
/lgtm

PR #357, opened 2024-10-16
Don't use advisory pipeline for staging component RPA

creydr:
/hold

Not entirely sure if we need this


creydr:
Won't change it, as advisories for staging go to a different location anyhow. See https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1729777953560109?thread_ts=1729771795.452959&cid=C04PZ7H0VA8

PR #358, opened 2024-10-16
[main] Add Konflux configurations

serverless-qe:
Add Konflux components and pipelines


creydr (approved):
/lgtm

RPA looks good
/unhold

https://github.com/openshift-knative/hack/pull/365 can be done in the next cycle

PR #365, opened 2024-10-18
Remove duplicate components in RAPs

creydr:
As per title


pkg/konfluxgen/konfluxgen.go
```diff
@@ -1012,6 +1012,7 @@ func getComponentImageRefs(csv *operatorsv1alpha1.ClusterServiceVersion) ([]Comp
 
 	soVersion := csv.Spec.Version.Version
 	componentVersion := soversion.ToUpstreamVersion(soVersion.String())
+	addedComponents := make(map[string]interface{})
```
pierDipi, line 1015: the best practice is to use `map[string]struct{}`

pierDipi (approved):
/lgtm
/approve

PR #372, opened 2024-10-21
Add cmd to get RHEL version of SO version

creydr:
Extracts the RHEL version of the SO config for a certain SO version.

### DIFF
--- a/pkg/rhel/rhel-version.go
+++ b/pkg/rhel/rhel-version.go
@@ -4,14 +4,25 @@ import (
 	"fmt"
 	"regexp"
 
+	"github.com/coreos/go-semver/semver"
+	"github.com/openshift-knative/hack/config"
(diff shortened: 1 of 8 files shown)


pkg/rhel/rhel-version.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/openshift-knative/hack/pkg/prowgen"
+)
+
+func ForSOBranchName(soBranchName string) (string, error) {
+	soConfig, loadErr := prowgen.LoadConfig("config/serverless-operator.yaml")
```
pierDipi, line 11: if we don't embed the file the path might not always be there

creydr: Uh. Yes. Updated

pierDipi (approved):
/lgtm

PR #374, opened 2024-10-22
Force Konflux Index image in custom weekly jobs

mgencur:
Forgot to change this in the previous PR.

### DIFF
--- a/config/serverless-operator.yaml
+++ b/config/serverless-operator.yaml
@@ -55,7 +55,7 @@ repositories:
           cluster_profile: aws
           test:
           - as: operator-e2e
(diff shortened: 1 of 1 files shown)


skonto:
@mgencur should we have a default somewhere or force it anyway for Serving too? See [here](https://github.com/openshift-knative/serving/pull/929#issuecomment-2430201540)

pierDipi:
@skonto for other midstream repos, the bundle and index need to be re-built otherwise the job will run against images that are configured on SO side, so jobs will not run on up-to-date components code and test that are specific to a feature on a specific branch will fail

PR #376, opened 2024-10-23
Use correct repo for bundle in RPA

creydr:
Follow up on https://redhat-internal.slack.com/archives/CKR568L8G/p1729684088850349 and use image repo `serverless-operator-bundle` for `serverless-bundle` component.


creydr:
@pierDipi @mgencur could you PTAL?

/cc @pierDipi @mgencur

PR #380, opened 2024-10-24
Konflux configurations architecture diagram

pierDipi:
As per title

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,17 @@ Limitations:
 
 - It is not currently possible to disable periodics per job
 
+# Konflux
+
+The architecture of the Konflux configuration architecture is summarized in this diagram:
+
+![Konflux applications and components diagram](./docs/assets/konflux-serverless-operator-architecture.png)
+
+while the following summarizes the Konflux release configuration architecture:
+
+![Konflux release diagram](./docs/assets/konflux-serverless-operator-architecture-release.png)
+
+
 ## Apply Konflux configurations
 
 1. Follow the instructions to access the Konflux instance


creydr (commented):
Nice! 
Where did you create those images? Is there a way, we could also add the "raw" file to this PR, so we could update the image later too?

pierDipi:
Raw parts are backed into the image, excalidraw is the tool

ReToCode:
Nice :) But 7.5 MB per Image in Git 🤷 ? Maybe cause of https://github.com/openshift-knative/hack/pull/380#issuecomment-2437041643? Or are there other things in it?

pierDipi:
> Nice :) But 7.5 MB per Image in Git 🤷 ? Maybe cause of [#380 (comment)](https://github.com/openshift-knative/hack/pull/380#issuecomment-2437041643)? Or are there other things in it?

No, there is only the image PNG + metadata to recreate the original design, what's the issue with that?

README.md
```diff
@@ -45,6 +45,17 @@ Limitations:
 
 - It is not currently possible to disable periodics per job
 
+# Konflux
+
+The architecture of the Konflux configuration architecture is summarized in this diagram:
```
skonto, line 50: Could you add some more info about design decisions made, so we may iterate on in the future.

ReToCode:
> No, there is only the image PNG + metadata to recreate the original design, what's the issue with that?

Seems oddly large for such a small image (draw.io) does the same and is in KB not MB. Especially if changed the git repo size grows for each revision - but anyway.

PR #381, opened 2024-10-25
SRVKE-1671: Ubi images have tzdata installed but zoneinfo is not there

pierDipi:
https: //access.redhat.com/solutions/5616681

### DIFF
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -241,8 +241,8 @@ func main() {
 				additionalInstructions: []string{},
 				"additional_packages":  strings.Join(additionalPackages, " "),
 			}
-			if strings.Contains(p, "ping") || slices.Contains(additionalPackages, "tzdata") {
-				// https: //access.redhat.com/solutions/5616681
+			if slices.Contains(additionalPackages, "tzdata") {
+				// https://access.redhat.com/solutions/5616681
 				d[additionalInstructions] = append(d[additionalInstructions].([]string), fmt.Sprintf("RUN microdnf update tzdata -y && microdnf reinstall tzdata -y"))
 				idx := -1
 				for i, p := range additionalPackages {
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/debug"
+	"slices"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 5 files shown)


cmd/generate/generate.go
```diff
@@ -225,16 +226,37 @@ func main() {
...
+				"builder":              builderImage,
+				"version":              metadata.Project.Tag,
+				"project":              projectWithSep,
+				"project_dashcase":     projectDashCaseWithSep,
+				"component":            capitalize(p),
+				"component_dashcase":   dashcase(p),
+				additionalInstructions: []string{},
+				"additional_packages":  strings.Join(additionalPackages, " "),
+			}
+			if strings.Contains(p, "ping") || slices.Contains(additionalPackages, "tzdata") {
```
maschmid, line 244: What exactly is the meaning of the `strings.Contains(p, "ping")` ?

pierDipi: pingsource it's a cmd directory called `mtping`

cmd/generate/generate.go
```diff
@@ -225,16 +226,37 @@ func main() {
...
+				// https: //access.redhat.com/solutions/5616681
```
creydr, line 245:
```suggestion
				// https://access.redhat.com/solutions/5616681
```

cmd/generate/generate.go
```diff
@@ -225,16 +226,37 @@ func main() {
...
+						idx = i
+						break
+					}
+				}
+				if idx >= 0 {
+					additionalPackages = slices.Delete(additionalPackages, idx, idx+1)
+				}
+			}
+			if len(additionalPackages) > 0 {
+				d[additionalInstructions] = append(d[additionalInstructions].([]string), fmt.Sprintf("RUN microdnf install %s", strings.Join(additionalPackages, " ")))
```
creydr, line 259:
```suggestion
				d[additionalInstructions] = append(d[additionalInstructions].([]string), fmt.Sprintf("RUN microdnf install -y %s", strings.Join(additionalPackages, " ")))
```

cmd/generate/generate.go
maschmid: ok, so this would only be added into the mtping image?  (I think the eventing-webhook might need it as well.)

cmd/generate/generate.go
pierDipi: no we don't that's the same as it was earlier

cmd/generate/generate.go
pierDipi: Good point, I'll add it through flag and remove this special casing here

creydr (approved):
/lgtm

pierDipi:
@creydr would you mind taking care of using this in Eventing ?

PR #368, opened 2024-10-21
Add new midstream branches and update pipelines

serverless-qe:
Discover new branches and update Konflux pipelines using openshift-knative/hack.

PR #384, opened 2024-10-25
Fix discovery to not update existing konflux configs

creydr:
This should help on https://github.com/openshift-knative/hack/pull/368#discussion_r1814799608


pkg/discover/discover.go
```diff
@@ -113,6 +113,11 @@ func discover(ctx context.Context, path string) error {
 								Enabled: true,
 							}
 						} else {
+							// copy the Konflux attribute as this is a pointer, and
+							// it would otherwise update the existing branch config
+							v := *other.Konflux
+							other.Konflux = &v
```
mgencur, line 119: Perhaps we should do the same with `other.Prowgen` here? and also on line https://github.com/openshift-knative/hack/pull/384/files#diff-493c512c6574cbf0c3e3e1a06288a2f10b8f8124cb27b95a29cf8ef4e48b6f0fR139 ?

pierDipi: or maybe worth looking at some deep copy libraries?

creydr: :+1: 
updated

creydr:
@mgencur @pierDipi could you recheck?

pkg/discover/discover.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/jinzhu/copier"
```
pierDipi, line 8: sort import

pierDipi (approved):
/lgtm

PR #386, opened 2024-10-29
Update Serving CI config for 1.16

skonto:
- Similar to https://github.com/openshift-knative/hack/pull/192
- 4.12 is EOL Jan/17/2025 so keeping it for now with 1.12.
- We don't have a 4.17 cluster-pool, probably we should?


ReToCode (approved):
/lgtm
/approve

/unhold (net-* branches are there now)

Not sure about the test errors though, cc @pierDipi ?

skonto:
@ReToCode could you stamp again?

PR #388, opened 2024-10-30
SRVCLI-401: :gift: Add kn-event configuration

cardil:
## Changes

 - :gift: Add kn-event configuration

Fixes [SRVCLI-401](https://issues.redhat.com/browse/SRVCLI-401)

Supersedes openshift-knative/hack#154
Requires openshift-knative/kn-plugin-event#540
Requires openshift-knative/kn-plugin-event#542
Requires openshift-knative/kn-plugin-event#543

### DIFF
--- a/.github/workflows/release-generate-ci.yaml
+++ b/.github/workflows/release-generate-ci.yaml
@@ -509,6 +509,30 @@ jobs:
           GH_TOKEN: ${{ secrets.SERVERLESS_QE_ROBOT }}
           GITHUB_TOKEN: ${{ secrets.SERVERLESS_QE_ROBOT }}
         if: ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule') && github.ref_name == 'main' }}
+        name: '[kn-plugin-func - release-v1.16] Create Konflux PR'
+        run: |
+          set -x
+          repo="kn-plugin-func"
+          branch="sync-konflux-release-v1.16"
+          target_branch="release-v1.16"
+          git remote add fork "https://github.com/serverless-qe/$repo.git" || true # ignore: already exists errors
+          remote_exists=$(git ls-remote --heads fork "$branch")
+          if [ -z "$remote_exists" ]; then
+            # remote doesn't exist.
+            git push "https://serverless-qe:${GH_TOKEN}@github.com/serverless-qe/$repo.git" "$branch:$branch" -f || exit 1
+          fi
+          git fetch fork "$branch"
+          if git diff --quiet "fork/$branch" "$branch"; then
+            echo "Branches are identical. No need to force push."
+          else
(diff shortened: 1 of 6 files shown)


config/kn-plugin-event.yaml
```diff
@@ -1,17 +1,17 @@
...
         enabled: true
       openShiftVersions:
         - version: '4.16'
           useClusterPool: true
+    release-1.16:
+      konflux:
+        enabled: true
+      openShiftVersions:
+        - version: '4.17'
+          useClusterPool: true
```
mgencur, line 14: AFAIK, we don't have a cluster pool for 4.17 yet. The version here should be 4.16 for now. When we bump the cluster pool we can increase the version in this place and in other repository configs as well.

cardil: I backed down it to 4.16 cluster pool in 02e43dad1961425861c9406246f1d19f2d4ee911 for now.

.github/workflows/release-generate-ci.yaml
```diff
@@ -430,12 +437,12 @@ jobs:
...
-        name: '[kn-plugin-func - release-v1.15] Create Konflux PR'
+        name: '[kn-plugin-event - release-1.15] Create Konflux PR'
         run: |
           set -x
-          repo="kn-plugin-func"
-          branch="sync-konflux-release-v1.15"
-          target_branch="release-v1.15"
+          repo="kn-plugin-event"
+          branch="sync-konflux-release-1.15"
+          target_branch="release-1.15"
```
cardil, line 445: This looks weird. The func config shouldn't be replaced by the event one :confused:.

Makefile
```diff
@@ -35,6 +35,10 @@ generate-functions-ci:
 	go run github.com/openshift-knative/hack/cmd/prowgen --config config/kn-plugin-func.yaml $(ARGS)
 .PHONY: generate-functions-ci
 
+generate-event-ci:
```
mgencur, line 38: Can you rename this to `generate-kn-plugin-event-ci` ? We already have `generate-eventing-ci`. It's confusing

.github/workflows/release-generate-ci.yaml
cardil: It looks like it was fixed in 2b5bdc51dcd3c954290f1f88772fa4ae33d4c6a4

Something must have been confused while resolving the conflict.

Makefile
cardil: The func plugin got `generate-functions-ci`, so I just followed that.

Maybe, we can change the function's target as well to have some consistency: `make generate-plugin-func-ci`

The `make generate-kn-plugin-event-ci` is so mouthful!

BTW. this convention is stuttering - `make generate-`. Better would be `make event-plugin`, `make eventing` etc.

mgencur: At minimum let's just rename your new target to generate-plugin-event-ci

PR #392, opened 2024-11-04
Use build args in FBC pipeline too

creydr:
We are getting a `Base image "registry.ci.openshift.org/origin/4.17" is from a disallowed registry` on the FBC applications:
image

This is, because we don't owerride the `ARG OPM_IMAGE=registry.ci.openshift.org/origin/4.17:operator-registry` arg in the index builds


pierDipi (approved):
/lgtm
/approve

PR #393, opened 2024-11-04
Use fbc-standard policy for FBC enterprise contract

creydr:
We should use the an EC Policy config for FBCs, when validating FBC apps:

image


pkg/konfluxgen/konfluxgen.go
```diff
@@ -324,6 +326,12 @@ func Generate(cfg Config) error {
 				r.Hermetic = "false"
 			}
 
+			if cfg.ECPolicyConfigName == "" {
+				r.ECPolicyConfiguration = "rhtap-releng-tenant/tmp-onboard-policy"
```
creydr, line 330: I've created https://issues.redhat.com/browse/SRVCOM-3391 as a follow up on this policy

PR #401, opened 2024-11-05
Switch to fbc-stage policy for EC of FBC applications

creydr:
Switch to [fbc-stage policy](https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/common/product/EnterpriseContractPolicy/fbc-stage.yaml) as discussed, as this excludes the `fbc-related-image-check`, which fails as the image references in the bundle point to registry.redhat.io and don't exist at that point yet.


mgencur:
Yeah. I think it really makes sense to do this. We will be testing the FBC image downstream using the ImageContentSourcePolicy. That one is generated so it will ensure that the images exist in quay.io. 
The fbc-standard policy really makes sense only for final push to production because at that point, the regular product images (other than fbc) should already be public under registry.redhat.io.

pierDipi (approved):
/lgtm

PR #406, opened 2024-11-06
update must gather default base image

Kaustubh-pande:
As build is [failing](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift-knative_serverless-operator/2990/pull-ci-openshift-knative-serverless-operator-main-416-images/1853857018083807232) due to base image, 

- update the base image
- override default value of must gather base image and as per https://redhat-internal.slack.com/archives/CKR568L8G/p1730883580226309 we are going to use Min OCP version Tag.
- Nit: add writeRPMLockFile function

cc @creydr

### DIFF
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -83,7 +83,6 @@ func main() {
 		dockerfilesSourceDir         string
 		projectFilePath              string
 		dockerfileImageBuilderFmt    string
-		mustGatherBuilderImageFmt    string
 		appFileFmt                   string
 		registryImageFmt             string
 		imagesFromRepositories       []string
@@ -113,7 +112,6 @@ func main() {
 	pflag.StringVar(&output, "output", filepath.Join(wd, "openshift"), "Output directory")
 	pflag.StringVar(&projectFilePath, "project-file", filepath.Join(wd, "openshift", "project.yaml"), "Project metadata file path")
 	pflag.StringVar(&dockerfileImageBuilderFmt, "dockerfile-image-builder-fmt", builderImageFmt, "Dockerfile image builder format")
-	pflag.StringVar(&mustGatherBuilderImageFmt, "must-gather-builder-image-fmt", mustGatherBaseImageFmt, "Must gather Dockerfile image builder format")
 	pflag.StringVar(&appFileFmt, "app-file-fmt", "/usr/bin/%s", "Target application binary path format")
 	pflag.StringVar(&registryImageFmt, "registry-image-fmt", "registry.ci.openshift.org/openshift/%s:%s", "Container registry image format")
 	pflag.StringArrayVar(&imagesFromRepositories, "images-from", nil, "Additional image references to be pulled from other midstream repositories matching the tag in project.yaml")
@@ -354,12 +352,7 @@ func main() {
 			log.Println("File ", projectFilePath, " not found")
 			metadata = nil
 		}
-		// Builder image might be provided without formatting '%s' string as plain value
-		builderImage := mustGatherBuilderImageFmt
-		if strings.Count(mustGatherBuilderImageFmt, "%s") == 1 {
-			builderImage = fmt.Sprintf(mustGatherBaseImageFmt, metadata.Requirements.OcpVersion.Max)
-		}
-
+		builderImage := fmt.Sprintf(mustGatherBaseImageFmt, metadata.Requirements.OcpVersion.Max)
 		projectName := mustGatherDockerfileTemplateName
 		projectDashCaseWithSep := projectName + "-"
 
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -36,7 +36,7 @@ const (
 	defaultDockerfileTemplateName      = "default"
 	funcUtilDockerfileTemplateName     = "func-util"
 	mustGatherDockerfileTemplateName   = "must-gather"
-	mustGatherBaseImage                = "brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:latest"
+	mustGatherBaseImageFmt             = "quay.io/openshift/origin-must-gather:%s"
 	// builderImageFmt defines the default pattern for the builder image.
 	// At the given places, the Go version from the projects go.mod will be inserted.
 	// Keep in mind to also update the tools image in the ImageBuilderDockerfile, when the OCP / RHEL
@@ -83,6 +83,7 @@ func main() {
 		dockerfilesSourceDir         string
 		projectFilePath              string
 		dockerfileImageBuilderFmt    string
+		mustGatherBuilderImageFmt    string
 		appFileFmt                   string
 		registryImageFmt             string
 		imagesFromRepositories       []string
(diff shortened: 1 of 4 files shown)


cmd/generate/generate.go
```diff
@@ -364,34 +354,26 @@ func main() {
 			log.Println("File ", projectFilePath, " not found")
 			metadata = nil
 		}
+		// Builder image might be provided without formatting '%s' string as plain value
+		builderImage := mustGatherBuilderImageFmt
+		if strings.Count(mustGatherBuilderImageFmt, "%s") == 1 {
```
mgencur, line 359: Cool. So this will generate the right Dockerfile for OpenShift CI where brew.registry.redhat.io is not available.
But we should still use the brew variant for Konflux build. We need to pass the build arg similar to other image overrides: https://github.com/openshift-knative/hack/blob/main/pkg/prowgen/prowgen_konflux.go#L100

mgencur: I believe we can just place it [here](https://github.com/openshift-knative/hack/blob/main/config/serverless-operator.yaml#L8) (and further below for release-1.35 branch) like this:
```
        - name: MUSTGATHER
          pullSpec: brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:latest
```

cmd/generate/generate.go
```diff
@@ -112,6 +113,7 @@ func main() {
 	pflag.StringVar(&output, "output", filepath.Join(wd, "openshift"), "Output directory")
 	pflag.StringVar(&projectFilePath, "project-file", filepath.Join(wd, "openshift", "project.yaml"), "Project metadata file path")
 	pflag.StringVar(&dockerfileImageBuilderFmt, "dockerfile-image-builder-fmt", builderImageFmt, "Dockerfile image builder format")
+	pflag.StringVar(&mustGatherBuilderImageFmt, "must-gather-builder-image-fmt", mustGatherBaseImageFmt, "Must gather Dockerfile image builder format")
```
creydr, line 116: Not sure if we need a new param for this, or if we could simply default to `quay.io/openshift/origin-must-gather:%s`

cmd/generate/generate.go
creydr: :+1: 
Only one point from my side: `latest` seems to be pretty old (`2021-11-18T17:17:43.151324972Z`). I guess we have to use the correct tag here (e.g. `brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:v4.17`)

creydr (commented):
Thanks for working on this @Kaustubh-pande

cmd/generate/generate.go
mgencur: For serverless 1.34 we were using this base image: `FROM openshift-ose-must-gather:v4.13.0 AS builder`. So, perhaps we want the oldest version from the list of supported ones.

mgencur: brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:v4.13 is upd-to-date: `"build-date": "2024-11-05T07:55:36"`

cmd/generate/generate.go
```diff
@@ -354,12 +352,7 @@ func main() {
 			log.Println("File ", projectFilePath, " not found")
 			metadata = nil
 		}
-		// Builder image might be provided without formatting '%s' string as plain value
-		builderImage := mustGatherBuilderImageFmt
-		if strings.Count(mustGatherBuilderImageFmt, "%s") == 1 {
-			builderImage = fmt.Sprintf(mustGatherBaseImageFmt, metadata.Requirements.OcpVersion.Max)
-		}
-
+		builderImage := fmt.Sprintf(mustGatherBaseImageFmt, metadata.Requirements.OcpVersion.Max)
```
creydr, line 355:
```suggestion
		builderImage := fmt.Sprintf(mustGatherBaseImageFmt, metadata.Requirements.OcpVersion.Min)
```
I guess @mgencur s suggestion in https://github.com/openshift-knative/hack/pull/406#discussion_r1830549938 was to use `Min`.

config/serverless-operator.yaml
```diff
@@ -14,6 +14,8 @@ config:
           pullSpec: registry.access.redhat.com/ubi8/openjdk-21
         - name: JAVA_RUNTIME
           pullSpec: registry.access.redhat.com/ubi8/openjdk-21-runtime
+        - name: MUSTGATHER
+          pullSpec: brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:latest
```
creydr, line 18: As we probably want to use the same version as in https://github.com/openshift-knative/hack/blob/b3ca9b30c790c15ab6105386155c73a3d140de54/cmd/generate/generate.go#L355, we probably need to add this dynamically with the pattern `brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:v + metadata.Requirements.OcpVersion.Min`.
You can probably do this here: https://github.com/openshift-knative/hack/blob/b3ca9b30c790c15ab6105386155c73a3d140de54/pkg/prowgen/prowgen_konflux.go#L269

cmd/generate/generate.go
Kaustubh-pande: here i guess we used either latest or Max bcz the default quay.io seems like not so updated with oldest tag, for 4.13 build 8 months ago,
https://quay.io/repository/openshift/origin-must-gather?tab=tags

Kaustubh-pande: here i guess we used either latest or Max bcz the default quay.io seems like not so updated with oldest tag, for 4.13 build 8 months ago,
https://quay.io/repository/openshift/origin-must-gather?tab=tags

Kaustubh-pande: martin comment is for `brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:v4.13` we can change it to Min instead of latest here https://github.com/openshift-knative/hack/pull/406/files#diff-8c104313544b2c9a8413d2cbb7a399e56793ffab4a3e554d5ac2243a5a74f1c8R18

config/serverless-operator.yaml
mgencur: Right, yeah

cmd/generate/generate.go
creydr: I think we can align those, so both tags are for the same OCP version.

pkg/prowgen/prowgen_konflux.go
```diff
@@ -75,16 +81,11 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
...
-							"olm-catalog", "serverless-operator", "project.yaml")
-						soMetadata, err := project.ReadMetadataFile(soProjectYamlPath)
-						if err != nil {
-							return err
-						}
 						versionLabel = soMetadata.Project.Version
 					}
 					log.Println("Version label:", versionLabel)
 					buildArgs = append(buildArgs, fmt.Sprintf("VERSION=%s", versionLabel))
+					buildArgs = append(buildArgs, fmt.Sprintf("MUSTGATHER=brew.registry.redhat.io/rh-osbs/openshift-ose-must-gather:v%s", soMetadata.Requirements.OcpVersion.Min))
```
creydr, line 88: as we need the mustgather arg only for SO builds, I think we can remove it here (have it in `GenerateKonfluxServerlessOperator`)

creydr (commented):
/lgtm

pkg/prowgen/prowgen_konflux.go
```diff
@@ -75,12 +81,6 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
 						}
 						// For non-existent branches we keep going and use downstreamVersion for versionLabel.
 					} else {
-						soProjectYamlPath := filepath.Join(soRepo.RepositoryDirectory(),
```
mgencur, line 78: Sorry, too late. project.ReadMetadataFile needs to be done after running GitCheckout., and it won't be available if the branch doesn't exist. I'm not sure this change will work.

mgencur: @Kaustubh-pande Why did you move this up? I don't get it.

creydr: I guess this was a leftover from https://github.com/openshift-knative/hack/pull/406#discussion_r1830927550. I'll submit a PR with the fix

creydr: https://github.com/openshift-knative/hack/pull/409

PR #408, opened 2024-11-06
Add support for WORKDIR in general dockerfile template

dsimansk:
Adds support to define `WORKDIR` as an optional stanza. 

I've formatted the template to prevent updates with newline in the existing Dockerfiles. Pls let me know if that's ok or can be any prettier wrt/ `text/template` placeholders. Otherwise it's working as expected not adding newline and changes if flag is not set. 

 

/cc @pierDipi @creydr

### DIFF
--- a/cmd/generate/dockerfile-templates/DefaultDockerfile.template
+++ b/cmd/generate/dockerfile-templates/DefaultDockerfile.template
@@ -3,12 +3,10 @@ ARG GO_BUILDER={{.builder}}
 ARG GO_RUNTIME=registry.access.redhat.com/ubi8/ubi-minimal
 
 FROM $GO_BUILDER as builder
-{{if .workdir}}
+
 WORKDIR /go/src/{{.main}}
 COPY . .
-{{else}}
-COPY . .
-{{end}}
+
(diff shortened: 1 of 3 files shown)


cmd/generate/dockerfile-templates/DefaultDockerfile.template
```diff
@@ -3,9 +3,12 @@ ARG GO_BUILDER={{.builder}}
...
 
 FROM $GO_BUILDER as builder
-
+{{if .workdir}}
+WORKDIR /go/src/{{.main}}
 COPY . .
-
+{{else}}
+COPY . .
+{{end}}
```
pierDipi, line 11: why not having it for all, without flags?

pierDipi: so that we start from a clean working directory and less potential issues?

dsimansk: OK, let me update the PR then.

pierDipi (approved):
/lgtm
/approve

PR #410, opened 2024-11-06
:lady_beetle: [kn-event] Proper unit tests target

cardil:
## Changes

- :lady_beetle: Proper Makefile target for unit tests for kn-event

See the target: 

https://github.com/openshift-knative/kn-plugin-event/blob/41fbdb9642d1ee516af1043a8efa4ee523656904/Makefile#L14-L16

### DIFF
--- a/config/kn-plugin-event.yaml
+++ b/config/kn-plugin-event.yaml
@@ -43,6 +43,6 @@ repositories:
   slackChannel: '#knative-client'
   tests:
   - as: unit
-    commands: make test
+    commands: make unit
     container:
       from: src


dsimansk:
As an example, we have deferred unit tests in S-O to gh actions. It's actually super quick to execute in comparison to OpenShift CI. Just an ides how to approach it from different side.

cardil:
The unit tests on Openshift CI are also fairly quick. They don't take much resources and time, and are executed in parallel with e2e tests, so they aren't the real problem IMHO.

PR #418, opened 2024-11-08
[main] Add Konflux configurations

serverless-qe:
Add Konflux components and pipelines


mgencur:
/lgtm

We need this for proper building of FBC catalog for OCP 4.18

PR #423, opened 2024-11-13
Do NOT add OCP variant to custom config names

mgencur:
I have spoken with INTEROP team and they rely on some automation and processes that need certain naming conventions. 
* Change `lp-rosa-classic` to `ocp4.17-lp-rosa-classic`
* As part of this change - prevent adding the OpenShift variant to the custom config name. It makes sense to hard code the name of the file/variant, i.e. take it directly from the config name. The custom configs often include OCP-specific configuration elements so they're bound to certain OCP version anyway.

### DIFF
--- a/config/serverless-operator.yaml
+++ b/config/serverless-operator.yaml
@@ -46,7 +46,7 @@ config:
         disabled: true
 repositories:
 - customConfigs:
-  - name: aws-ovn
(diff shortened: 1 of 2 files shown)


ReToCode (approved):
/lgtm
/approve

PR #424, opened 2024-11-13
Add repos LICENSE file in image under /licenses dir

creydr:
Adding the repos LICENSE file to the image under /license.

See https://redhat-internal.slack.com/archives/CKR568L8G/p1731535620986049?thread_ts=1731455170.342769&cid=CKR568L8G and https://redhat-internal.slack.com/archives/CKR568L8G/p1731915836475649 for full context.

### DIFF
--- a/cmd/generate/dockerfile-templates/FuncUtilDockerfile.template
+++ b/cmd/generate/dockerfile-templates/FuncUtilDockerfile.template
@@ -4,7 +4,7 @@ ARG GO_RUNTIME=registry.access.redhat.com/ubi8/ubi-minimal
 
 FROM $GO_BUILDER as builder
 
-WORKDIR /go/src/{{.main}}
+WORKDIR /workspace
 COPY . .
 
 ENV CGO_ENABLED=1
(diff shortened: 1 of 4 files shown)


cmd/generate/dockerfile-templates/FuncUtilDockerfile.template
```diff
@@ -4,6 +4,7 @@ ARG GO_RUNTIME=registry.access.redhat.com/ubi8/ubi-minimal
 
 FROM $GO_BUILDER as builder
 
+WORKDIR /go/src/{{.main}}
```
matejvasek, line 7: Couldn't we have just `WORKDIR /workspace` for sake of simplicity? I think that for long time project does not have to be in `$GOPATH/src`. 

And even if we wanted to use the old way then the path should **not** be `/go/src/{{.main}}` (evaluated to `/go/src/cmd/func-util`).
It should be more something like `/go/src/{{.module}}` (evaluated to something like `/go/src/github.com/knative/func`).

matejvasek: `/go/src/{{.main}}` just doesn't seems right even thought it might work.

matejvasek: See [GOPATH development mode](https://go.dev/wiki/GOPATH#gopath-development-mode) vs [Deprecating and removing GOPATH development mode](https://go.dev/wiki/GOPATH#deprecating-and-removing-gopath-development-mode).

dsimansk: It can be pretty much anything. For the issue with multi-arch builds, the only problem was staying on root `/`. However, it's only workdir. There are no other build consequence AFAIK. The convenience was that we had `{{.main}}` already present in the template. It can be whatever, but to it's a build impl detail that doesn't matter unless it's breaking anything.

creydr: updated

PR #446, opened 2024-11-25
Reflect changes from hack repo in konflux-apply quicker

creydr:
Update konflux-apply to apply changes from hack repo before and after the other components so that it gets reflected quickly.


pierDipi (approved):
/lgtm
/approve

PR #453, opened 2024-11-27
Fix listing branches to include current highlighted branch

pierDipi:
In `git --no-pager branch --list "release-*"`, the current branch is highlighted with a star and the parser is not handling that case:

```
$ git --no-pager branch --list "release-*"
  release-1.33
  release-1.34
* release-1.35
```


creydr (approved):
/lgtm

PR #459, opened 2024-11-28
[release-v1.15] Add signed image override to client config

dsimansk:
/cc @creydr @pierDipi @mgencur

### DIFF
--- a/config/client.yaml
+++ b/config/client.yaml
@@ -3,6 +3,9 @@ config:
     release-v1.15:
       konflux:
         enabled: true
+      imageOverrides:
+        - name: CLI_ARTIFACTS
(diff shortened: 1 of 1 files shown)


config/client.yaml
```diff
@@ -3,6 +3,9 @@ config:
     release-v1.15:
       konflux:
         enabled: true
+      imageOverrides:
+      - name: CLI_ARTIFACTS
+        pullSpec: brew.registry.redhat.io/rh-osbs/openshift-serverless-1-kn-cli-artifacts-rhel8:1.15.0
```
creydr, line 8: @mgencur do you remember the list of allowed registries for Konflux?

dsimansk: My understanding was going through ARGS and override it on the pipeline is preferred approach per Martin's suggestion here: https://github.com/openshift-knative/client/pull/525#discussion_r1840659335

creydr (approved):
/lgtm

PR #461, opened 2024-11-29
Don't run SO pipelines on `.konflux-release` dir updates

creydr:
Don't run SO pipelines on `.konflux-release` dir updates


pierDipi (approved):
/lgtm
/approve

PR #463, opened 2024-12-02
Specify 4.18 as on-demand for release-1.35 and release-1.36

mgencur:
This should be OK next time when "discover branches" runs. We have recently fixed the logic and the openshift versions are copied from the "latest" config in the file.


pierDipi (approved):
/lgtm

PR #464, opened 2024-12-04
konfluxgen: Create IntegrationTestScenarios for override snapshots

creydr:
Update konfluxgen to create IntegrationTestScenarios for the override snapshots with the prod policies, so we get early feedback on EC runs

### DIFF
--- a/pkg/konfluxgen/integration-test-scenario.template.yaml
+++ b/pkg/konfluxgen/integration-test-scenario.template.yaml
@@ -1,7 +1,7 @@
 apiVersion: appstudio.redhat.com/v1beta2
 kind: IntegrationTestScenario
 metadata:
-  name: {{{ truncate ( sanitize .ApplicationName ) }}}-enterprise-contract
+  name: {{{ truncate ( sanitize .Name ) }}}
 spec:
   params: 
     - name: POLICY_CONFIGURATION
@@ -10,8 +10,10 @@ spec:
       value: "120m"
   application: {{{ truncate ( sanitize .ApplicationName ) }}}
   contexts:
-    - description: Application testing
-      name: application
+    {{{- range .Contexts }}}
+    - description: {{{ .Description }}}
+      name: {{{ .Name }}}
+    {{{- end }}}
   resolverRef:
     params:
       - name: url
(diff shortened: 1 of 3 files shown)


pkg/konfluxgen/integration-test-scenario.template.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: appstudio.redhat.com/v1beta2
 kind: IntegrationTestScenario
 metadata:
-  name: {{{ truncate ( sanitize .ApplicationName ) }}}-enterprise-contract
```
creydr, line 4: as the new ITC do not have the `-enterprise-contract` suffix, we need to delete the old ones, when this merged and got applied

pierDipi: it helps to know what's the test about from the name

pkg/konfluxgen/konfluxgen.go
```diff
@@ -448,6 +426,65 @@ func Generate(cfg Config) error {
...
+		}
+		if err := WriteFileReplacingNewerTaskImages(filepath.Join(ecTestDir, "ec-test.yaml"), buf.Bytes(), 0777); err != nil {
+			return fmt.Errorf("failed to write application file: %w", err)
+		}
+
+		buf.Reset()
+
+		// add integration test scenario for override snapshots with prod policies
+		config = IntegrationTestConfig{
+			Name:            fmt.Sprintf("%s-override-snapshot", appKey),
```
pierDipi, line 464: kinda related to the above comment, I miss an EC reference in the name

creydr: wenn we have the EC reference in the name, then the name will end up with something like `serverless-operator-135-override-snapshot-enterprise-contract` and will be hashed. That was also the reason, why I removed the suffix it above.

pierDipi: what about short names? `ec` or `ovr` ?

pierDipi: like `serverless-operator-135-override-snapshot-ec` or `serverless-operator-135-ec` ?

pierDipi (approved):
/lgtm
/approve

PR #465, opened 2024-12-05
Multi arch builds for FBC

rudyredhat1:
Hi all,

Ticket link - https://issues.redhat.com/browse/SRVCOM-3460 
Slack thread link - https://redhat-internal.slack.com/archives/CKR568L8G/p1733204792164679

@creydr I have tried the changes that you have said. Rest can you all please help with the changes.

Thanks
@pierDipi @mgencur


pkg/konfluxgen/kustomize/fbc-builder.yaml
```diff
@@ -324,11 +312,7 @@ spec:
       operator: in
       values:
       - "false"
-    workspaces:
-    - name: workspace
-      workspace: workspace
   workspaces:
-  - name: workspace
```
pierDipi, line 331: Since this is not needed anymore, we can remove this section from the PipelineRun that references this:

https://github.com/openshift-knative/hack/blob/5d3194a5542e6106601535b543d4c625624504e1/pkg/konfluxgen/pipeline-run.template.yaml#L84-L96

pkg/konfluxgen/kustomize/fbc-builder.yaml
```diff
@@ -94,10 +72,24 @@ spec:
...
+  - default: ""
+    description: Path to a file with build arguments for buildah, see https://www.mankier.com/1/buildah-build#--build-arg-file
+    name: build-args-file
+    type: string
+  - default:
+    - linux/x86_64
+    description: List of platforms to build the container images on. The available
+      set of values is determined by the configuration of the multi-platform-controller.
+    name: build-platforms
+    type: array
```
pierDipi, line 92: We would need to pass this parameter in the PipelineRun:

https://github.com/openshift-knative/hack/blob/main/pkg/konfluxgen/pipeline-run.template.yaml#L49-L56

so the check would become something like this:

```
    - name: build-platforms
      value:
        - linux/x86_64
    {{{- if eq .Pipeline "fbc-builder" }}}
    - name: skip-fbc-related-image-check
      value: "true"
    {{{- end }}}
```

creydr: Good point

pkg/konfluxgen/docker-build.yaml
```diff
@@ -420,7 +420,113 @@ spec:
...
+      - name: bundle
+        value: quay.io/konflux-ci/tekton-catalog/task-coverity-availability-check-oci-ta:0.1@sha256:df12720aae4edaba04d374bac7df2990746821821ef426ad7c0b2c65e473d458
+      - name: kind
+        value: task
+      resolver: bundles
+    when:
+    - input: $(params.skip-checks)
+      operator: in
+      values:
+      - "false"
```
pierDipi, line 487: `tasks.build-container` doesn't exist in this pipeline

pierDipi: see here https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1733475095120619?thread_ts=1733475087.909069&cid=C04PZ7H0VA8

for now, we can remove these 2 tasks

pkg/konfluxgen/fbc-builder.yaml
```diff
@@ -125,61 +116,16 @@ spec:
     value: $(tasks.clone-repository.results.commit)
   tasks:
   - name: fbc-related-image-check
-    runAfter:
```
mgencur, line 128: How can this work? Just listing a task without referencing `taskRef`.
Same for the build-container task below. Was this removed by accident?

If I understand it correctly this "fbc-related-image-check" task is now merged into "validate-fbc" task that you're adding in this PR. So we can probably remove this fbc-related-image-check.
Also, the parameter "skip-fbc-related-image-check" for skipping this task in pre-submit checks won't work. We can probably remove it completely because skipping the whole "validate-fbc" is probably not a good thing.

pkg/konfluxgen/pipeline-run.template.yaml
```diff
@@ -50,7 +50,7 @@ spec:
     - name: build-platforms
       value:
         - linux/x86_64
-    {{{- else }}}
+    {{{- if eq .Pipeline "fbc-builder" }}}
```
pierDipi, line 53: the if above needs to be removed

pkg/konfluxgen/fbc-builder.yaml
rudyredhat1: - name: fbc-related-image-check
    when:
    - input: $(params.skip-fbc-related-image-check)
      operator: in
      values:
      - "false"
      
Removing line 118 to 123 will work? @mgencur 
and rest of the changes are done, is there any change required @pierDipi @creydr

mgencur: > Removing line 118 to 123 will work?
It should work. I don't see how it can work in the current form. But we'll need to test!

pkg/konfluxgen/fbc-builder.yaml
```diff
@@ -115,12 +115,6 @@ spec:
     name: CHAINS-GIT_COMMIT
     value: $(tasks.clone-repository.results.commit)
   tasks:
-  - name: fbc-related-image-check
-    when:
-    - input: $(params.skip-fbc-related-image-check)
```
mgencur, line 140: The "skip-fbc-related-image-check" is now redundant and should be removed too.

pkg/konfluxgen/fbc-builder.yaml
```diff
@@ -115,12 +115,6 @@ spec:
...
-      operator: in
-      values:
-      - "false"
   - name: build-container
```
mgencur, line 118: Does `build-container` really work? It doesn' reference any task. I suppose it will fail.

pkg/konfluxgen/fbc-builder.yaml
rudyredhat1: This part to be removed right? L40-43
- default: "false"
    description: Skip fbc-related-image-check
    name: skip-fbc-related-image-check
    type: string

pkg/konfluxgen/fbc-builder.yaml
creydr: The `fbc-related-image-check` task seems not to be included in the new fbc multiarch pipeline. So I guess we can remove the `patch_related_image_check.patch`.
Also the `patch_build_args.patch` seems not to be needed anymore for the FBC, as we have this in the new FBC pipeline: https://github.com/openshift-knative/hack/blob/f69466d0e07e832521daf106a3ed1780fe8d67b2/pkg/konfluxgen/fbc-builder.yaml#L239-L241

pkg/konfluxgen/fbc-builder.yaml
mgencur: Yeah. And also passing the argument in the PipelineRun.

pkg/konfluxgen/fbc-builder.yaml
creydr: I guess we can remove the FBC patches: https://github.com/openshift-knative/hack/pull/465#discussion_r1873090669

pkg/konfluxgen/pipeline-run.template.yaml
```diff
@@ -46,11 +46,10 @@ spec:
...
     - name: output-image
       value: quay.io/redhat-user-workloads/ocp-serverless-tenant/{{{ truncate ( sanitize .ApplicationName ) }}}/{{{ truncate ( sanitize .ProjectDirectoryImageBuildStepConfiguration.To ) }}}:on-pr-{{revision}}
-    {{{- if ne .Pipeline "fbc-builder" }}}
     - name: build-platforms
       value:
         - linux/x86_64
-    {{{- else }}}
+    {{{- if eq .Pipeline "fbc-builder" }}}
     - name: skip-fbc-related-image-check
       value: "true"
```
mgencur, line 54: This should be removed as well as mentioned in https://github.com/openshift-knative/hack/pull/465/files#r1873091151

pkg/konfluxgen/pipeline-run.template.yaml
```diff
@@ -49,9 +49,6 @@ spec:
     - name: build-platforms
       value:
         - linux/x86_64
-    {{{- if eq .Pipeline "fbc-builder" }}}
-    - name: skip-fbc-related-image-check
-      value: "true"
     {{{- end }}}
```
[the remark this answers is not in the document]
rudyredhat1: one more doubt, but previously we were having the skip-fbc-related-image-check only for the fbc-builder, now we are completely removing that, right?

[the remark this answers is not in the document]
mgencur: That flag only makes sense for fbc-builder. The docker-build doesn't use that. We can remove it.

mgencur:
LGTM now but I will leave merging up to @pierDipi @creydr

PR #468, opened 2024-12-06
Add flag to extended builder env vars

dsimansk:
This should enable event sender or `kn` to be vendorless. 

There are a few other options, but the straight-forward env flag seemed the most practical to me. 

Other options:
- Inferring from existing or not-existing `./vendor` dir 
- Having `--vendorless` flag on the generator
- ... 

/cc @creydr @pierDipi @mgencur


dsimansk:
@cardil per our chat on Slack.

pierDipi (approved):
/lgtm
/approve

creydr:
maybe we could update the funcutils template too, to support those params

dsimansk:
> maybe we could update the funcutils template too, to support those params

Yep, good point. Thanks!

PR #470, opened 2024-12-06
Add PipelineRun specific tag

pierDipi:
We have an operator repository that is aggregating other components images using the custom `latest` tag which eventually turns into a resolved SHA.

When a pipeline runs on push for any "other components", a task after pushing or apply-tags could fail (for whatever reason), in that case, if we re-run the pipeline, the image revision (and therefore tag) is the same, but the SHA is different which ends up invalidating the "previous SHAs" in the operator repository.

Tested in https://github.com/openshift-knative/serverless-operator/pull/3141


creydr (approved):
/lgtm

PR #473, opened 2024-12-09
Release Build Status table

pierDipi:
Currently, there is no simple way to know what's the build status of all of the components

This is simple site, we can deploy with GH pages that aggregates build status of the last commit for all repositories we build in Konflux

Example view

image


creydr (approved):
/lgtm

PR #476, opened 2024-12-10
Group builds by build status

pierDipi:
this helps with identifying what to re-run


pierDipi:
I don't think it's worth it for now

pierDipi:
@creydr can we merge and see to do that in the future? I'm spending too much time reconciling the status of the prod builds

PR #479, opened 2024-12-11
Respect imageOverrides from all configs

creydr:
As per title to respect e.g. imageOverrides from client too

### DIFF
--- a/pkg/prowgen/prowgen_konflux.go
+++ b/pkg/prowgen/prowgen_konflux.go
@@ -103,11 +103,26 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
 							return fmt.Errorf("main or %s branch configuration not found for serverless-operator", soBranchName)
 						}
 					}
+
+					overrides := make(map[string]string)
+					// add overrides from SO config
 					for _, img := range br.Konflux.ImageOverrides {
 						if img.Name == "" || img.PullSpec == "" {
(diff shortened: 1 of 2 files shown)


dsimansk (approved):
/approve
/lgtm

pkg/prowgen/prowgen_konflux.go
```diff
@@ -103,11 +103,26 @@ func GenerateKonflux(ctx context.Context, openshiftRelease Repository, configs [
...
+					for _, img := range b.Konflux.ImageOverrides {
+						if img.Name == "" || img.PullSpec == "" {
+							return fmt.Errorf("image override missing name or pull spec: %#v", img)
+						}
+						overrides[img.Name] = img.PullSpec
+					}
+
+					for name, pullSpec := range overrides {
+						buildArgs = append(buildArgs, fmt.Sprintf("%s=%s", name, pullSpec))
 					}
```
pierDipi, line 126: why would we loop and create a separate variable when we can just aggregate on buildArgs ?

creydr: Because this would not override the existing ones. For example, when we have an imageOverride in the SO config and one in the client config with the same name, this would end up in having two overrides.

```
GO_BUILDER=<from SO config>
GO_BUILDER=<from other config>
```

creydr: I only wanted to sort out that case too :shrug:

PR #480, opened 2024-12-11
Remove FORCE_KONFLUX_INDEX flag

mgencur:
The flag doesn't work in all cases. For example, there can be changes in serverless-operator images (like knative-openshift) and matching changes in tests. However,  after merging that pull request, only the specific operator image will be built but the Konflux index will still point to old operator images. In this case, running with an out-of-date Konflux index makes the tests fail and people have to investigate these false alarms which might be tricky. The Konflux index will be ready only after a few "make generated-files" runs against s-o.
It is generally safer to always build the Bundle AND Index image within the CI run and have everything up to date. Forcing the Konflux index image makes sense only for specific builds when we're sure all changes propagated to the Index image properly.

### DIFF
--- a/config/serverless-operator.yaml
+++ b/config/serverless-operator.yaml
@@ -16,22 +16,19 @@ config:
           pullSpec: registry.access.redhat.com/ubi8/openjdk-21-runtime
       openShiftVersions:
       - candidateRelease: true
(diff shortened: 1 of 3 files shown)


pierDipi (approved):
Makes sense, it was a good try!
/lgtm
/approve

PR #483, opened 2024-12-11
Return max check-runs per page

creydr:
Currently we list only 30 checkruns per repo. This can lead to the issue, that some checkruns are not show on the dashboard, when the result is paginaged from GB. As a quick fix, this PR sets `per_page` to the max allowed 100.


pierDipi (approved):
/lgtm
/approve

PR #484, opened 2024-12-12
Allow setting jobTimeout explicitly

mgencur:
* If not specified, it is testTimeout + 1 hour. Even if the default 4h timeout is used for "test" step, there needs to be some buffer for running must-gather.
* Reduce test timeout for upgrade tests - successful runs take 50-60 minutes so limiting to 1h30m should be enough. 
* Set jobTimeout for upgrade tests explicitly.


cardil (approved):
/lgtm

PR #486, opened 2024-12-16
Gather test logs on failure

mgencur:
Depends on https://github.com/openshift-knative/serverless-operator/pull/3185

When a test container in CI is interrupted due to timeout or infrastructure failures the test log is lost. This change will gather logs from a shared dir where they're copied earlier by the "test" container (see the related PR)

### DIFF
--- a/pkg/prowgen/prowgen_tests.go
+++ b/pkg/prowgen/prowgen_tests.go
@@ -130,6 +130,22 @@ func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, sk
 					Post: []cioperatorapi.TestStep{
 						{
 							LiteralTestStep: &cioperatorapi.LiteralTestStep{
+								As:       "testlog-gather",
+								From:     sourceImageName,
+								Commands: `cp -v ${SHARED_DIR}/debuglog-*.log ${SHARED_DIR}/stdout-*.log ${SHARED_DIR}/stderr-*.log "${ARTIFACT_DIR}/" || true`,
+								Resources: cioperatorapi.ResourceRequirements{
+									Requests: cioperatorapi.ResourceList{
+										"cpu": "100m",
+									},
+								},
+								Timeout:           &prowapi.Duration{Duration: 1 * time.Minute},
+								BestEffort:        pointer.Bool(true),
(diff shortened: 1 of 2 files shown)


pkg/prowgen/prowgen_tests.go
```diff
@@ -128,6 +128,22 @@ func DiscoverTests(r Repository, openShift OpenShift, sourceImageName string, sk
 						},
 					},
 					Post: []cioperatorapi.TestStep{
+						{
+							LiteralTestStep: &cioperatorapi.LiteralTestStep{
+								As:       "testlog-gather",
+								From:     sourceImageName,
+								Commands: `cp -v ${SHARED_DIR}/debuglog-*.log ${SHARED_DIR}/stdout-*.log ${SHARED_DIR}/stderr-*.log "${ARTIFACT_DIR}/" || true`,
```
cardil, line 135: Maybe put those in a sub-dir to avoid clobber?

mgencur: Discussed on [Slack](https://redhat-internal.slack.com/archives/CLUCMK7R6/p1734345988932009)

cardil: Directory structure looks like this https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/59819/rehearse-59819-pull-ci-mgencur-serverless-operator-archive_debug_logs-417-operator-e2e/1867594409290240000/artifacts/operator-e2e/test-logs-gather/artifacts/

So, the change of overwriting some files is low.

Still, I would put those in a subdir (just in case), but not binding for that.

/lgtm
/approve
/hold

_(Please, unhold if you think this is not an issue)_

mgencur: The name of log file includes the name of the calling script (see [link](https://github.com/openshift-knative/serverless-operator/pull/3185/files#diff-00836bf8713db966a23eb10670aca0fa847c67c61522f213a0410961a5164c49R59)) so I think the chance of overwriting is minimal.

/unhold

PR #494, opened 2025-01-03
Update tzdata to tzdata-2024b-4

pierDipi:
`tzdata` is now at version `2024b-4` which causes the builds to fail for Eventing https://github.com/openshift-knative/eventing/pull/986/checks?check_run_id=35101189570 and CDN https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/ppc64le/baseos/os/Packages/t/


creydr:
Hi @pierDipi,
where did you get the information from to update (and create) the `rpms.lock.yaml` file?

PR #498, opened 2025-01-07
Fix: Use correct policy for override-snapshot EC

creydr:
I missed this fix in #488, when I had to add the `-prod` suffix to the policy :(


pierDipi (approved):
/lgtm

PR #504, opened 2025-01-13
Add Serving's schema-tweak image exclude

dsimansk:
Per our chat earlier. This is upstream addition, however it's util cmd that doesn't need image to build for it. 

/cc @skonto

### DIFF
--- a/config/serving.yaml
+++ b/config/serving.yaml
@@ -51,6 +51,8 @@ repositories:
     - openshift/ci-operator/knative-images.*
     - openshift/ci-operator/knative-test-images.*
     - openshift/ci-operator/knative-perf-images.*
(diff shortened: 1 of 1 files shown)


dsimansk:
At least 1 approval and perhaps a quick review, if the config works. Thanks!
/cc @pierDipi @creydr

skonto:
The command I used in the past to test this is:
```
make generate-ci-no-clean ARGS=--branch=master
```
Right now it does not run on my side due to bandwidth issues or some git client issues. I am hitting:
```
fatal: fetch-pack: invalid index-pack output
```
I can't recall how I resolved it in the past. In any case I think it should be fine let's merge and check.

PR #507, opened 2025-01-15
Run make generate-release periodically and on push

pierDipi:
This will prevent repositories to get out of sync with Dockerfile generate


creydr (approved):
/lgtm

PR #514, opened 2025-01-17
Add workflow to apply override snapshots and create release CR PR

creydr:
Adding a workflow which:
* applies the override-snapshot for a given revision
* creates a PR with the release CRs referencing this override-snapshot(s)


.github/workflows/apply-konflux-override-snapshot.yaml
```diff
@@ -0,0 +1,33 @@
+name: Apply Konflux Override Snapshot
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "SO branch with the manifests"
```
pierDipi, line 6:
```suggestion
      revision:
        description: "SO branch or revision with the manifests"
```

pierDipi: or `reference` since this will work with both branch or revision

pierDipi: but branch is ok too

.github/workflows/apply-konflux-override-snapshot.yaml
```diff
@@ -0,0 +1,33 @@
...
+    shell: bash
+jobs:
+  snapshot-apply:
+    name: Apply Konflux Override Snapshot
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout serverless operator
+        uses: actions/checkout@v4
+        with:
+          ref: "${{ inputs.branch }}"
```
creydr, line 20:
```suggestion
          ref: "${{ inputs.revision }}"
```

creydr:
@pierDipi can you recheck?

.github/workflows/generate-release-crs.yaml
```diff
@@ -0,0 +1,96 @@
...
+          - stage
+          - prod
+        required: true
+      type:
+        description: "Type to release"
+        type: choice
+        options:
+          - component
+          - FBC
+        required: true
```
pierDipi, line 22: can't we create both PRs at the same time?

.github/workflows/generate-release-crs.yaml
```diff
@@ -0,0 +1,96 @@
...
+        description: "SO branch or revision with the manifests"
+        default: "main"
+        required: true
+      environment:
+        description: "Environment to release to"
+        type: choice
+        options:
+          - stage
+          - prod
+        required: true
```
pierDipi, line 15:
```suggestion
        required: true
        default: "prod"
```

pierDipi: that's the most likely option

.github/workflows/generate-release-crs.yaml
```diff
@@ -0,0 +1,96 @@
+name: Apply Konflux Override Snapshot and create Release CR PR
+on:
+  workflow_dispatch:
+    inputs:
+      revision:
+        description: "SO branch or revision with the manifests"
+        default: "main"
```
pierDipi, line 7:
```suggestion
        default: "release-1."
```

maybe that's a better default so that I only need to type the minor version ?

.github/workflows/generate-release-crs.yaml
creydr: Good point. Updated it

Makefile
```diff
@@ -43,6 +43,10 @@ discover-branches: clean
 	go run github.com/openshift-knative/hack/cmd/discover $(ARGS)
 .PHONY: discover-branches
 
+generate-konflux-release: clean
+	go run github.com/openshift-knative/hack/cmd/konflux-release-gen $(ARGS)
+.PHONY: discover-branches
```
creydr, line 48:
```suggestion
.PHONY: generate-konflux-release
```

creydr:
Rebased to fix merge conflicts
@pierDipi can you recheck?

pkg/konfluxgen/konfluxgen.go
```diff
@@ -1192,3 +1210,110 @@ func consistentVersion(cfg Config, csv *operatorsv1alpha1.ClusterServiceVersion)
...
+		counter, err := strconv.Atoi(strCounter)
+		if err != nil {
+			return "", fmt.Errorf("could not parse suffix %q of %q to number", strCounter, releaseMetadata.Name)
+		}
+
+		return fmt.Sprintf("%s-%d", releasePlanName, counter+1), nil
+	} else {
+		return "", fmt.Errorf("last release name %q does not match pattern '%s-<counter>'", releaseMetadata.Name, releasePlanName)
+	}
+}
```
pierDipi, line 1304: this will always increment the counter even on the ones we don't really want/need to increase

pkg/konfluxgen/konfluxgen.go
```diff
@@ -1192,3 +1210,110 @@ func consistentVersion(cfg Config, csv *operatorsv1alpha1.ClusterServiceVersion)
...
+	}
+
+	return nil
+}
+
+func executeReleaseTemplate(data Release, outputFilePath string) error {
+	funcs := template.FuncMap{
+		"sanitize": Sanitize,
+		"truncate": Truncate,
+	}
```
pierDipi, line 1258: those are not really used in the template

pkg/konfluxgen/konfluxgen.go
pierDipi: I think it's ok to not have support for retries and do it manually / document how to do it

pierDipi: what we need to ensure is that this doesn't override existing ones if they have being modified for retrying

creydr:
/hold
as I want to test more

creydr:
/hold cancel
@pierDipi I updated it, that it checks in the cluster for an existing release and its status for the retry case

pierDipi (approved):
/lgtm

PR #516, opened 2025-01-20
Add Serverless Operator 1.35.0 component release CR

creydr:
Adding the component release CR for Serverless Operator 1.35.0 (override snapshot `serverless-operator-135-override-snapshot-zg96j`).

Stage release with the override snapshot:
```
$ k get release --sort-by=.metadata.creationTimestamp | grep -i serverless-operator-135-override-snapshot-zg96j
manual-stage-release-override-snapshot-zg96j-g2gg2              serverless-operator-135-override-snapshot-zg96j           serverless-operator-135-1350-stage           Failed           3d21h
manual-stage-release-override-snapshot-zg96j-xvwtn              serverless-operator-135-override-snapshot-zg96j           serverless-operator-135-1350-stage           Failed           3d7h
manual-stage-release-override-snapshot-zg96j-rtk9c              serverless-operator-135-override-snapshot-zg96j           serverless-operator-135-1350-stage           Failed           3d6h
manual-stage-release-override-snapshot-zg96j-cjj2n              serverless-operator-135-override-snapshot-zg96j           serverless-operator-135-1350-stage           Failed           3d3h
manual-stage-release-override-snapshot-zg96j-wz9qm              serverless-operator-135-override-snapshot-zg96j           serverless-operator-135-1350-stage           Failed           2d21h
manual-stage-release-override-snapshot-zg96j-zcbzz              serverless-operator-135-override-snapshot-zg96j           serverless-operator-135-1350-stage           Succeeded        2d6h
```

**Merging this will trigger a production release pipeline of the SO 1.35.0 components when #515 is in!**

/hold to wait for all approvals (QE, docs & P/Z)

PR #519, opened 2025-01-22
Add SO 1.35.0 component release CR (2nd run)

creydr:
This is some preparation in case the release pipeline of #516 fails as a release pipeline can (yet) only be retriggered with a new release CR.

/hold to trigger only in case run of #516 failed


cardil (approved):
/lgtm

This shouldn't be necessary, but apparently Konflux is forcing us to adopt such lovely practices!

PR #520, opened 2025-01-22
Add SO 1.35.0 component release CR (3rd run)

creydr:
This is some preparation in case the release pipeline of #519 fails as a release pipeline can (yet) only be retriggered with a new release CR.

/hold to trigger only in case run of #519 failed


creydr:
/hold cancel
as pipeline of #519 (https://console.redhat.com/application-pipeline/workspaces/rhtap-releng/applications/serverless-operator-135/pipelineruns/managed-6xvr7) failed too

cardil (approved):
/lgtm

Here we go again...

PR #522, opened 2025-01-23
Add SO 1.35.0 component release CR (4. run)

creydr:
This is some preparation in case the release pipeline of #520 fails as a release pipeline can (yet) only be retriggered with a new release CR.

/hold to trigger only in case run of #520 failed


cardil (approved):
/lgtm

4th time's the charm

creydr:
/hold cancel
as https://console.redhat.com/application-pipeline/workspaces/rhtap-releng/applications/serverless-operator-135/pipelineruns/managed-wzqrm from #520 failed

PR #523, opened 2025-01-23
Add SO 1.35.0 component release CR (5. run)

creydr:
Same as #516 but with a release CR with another name (same `.spec` as in #516).
This is some preparation in case the release pipeline of #522 fails as a release pipeline can (yet) only be retriggered with a new release CR.

/hold to trigger only in case run of #522 failed

PR #524, opened 2025-01-23
Add SO 1.35.0 component release CR (6. run)

creydr:
This is some preparation in case the release pipeline of #523 fails as a release pipeline can (yet) only be retriggered with a new release CR.

/hold to trigger only in case run of #523 failed


creydr:
Run from #523 (https://console.redhat.com/application-pipeline/workspaces/rhtap-releng/applications/serverless-operator-135/pipelineruns/managed-rhmwj) failed due to a timeout issue. Updated the release pipeline again.

pierDipi (approved):
/lgtm
/approve

PR #526, opened 2025-01-24
Fix RP and RPA name / version during release cut 

pierDipi:
We use the csv version but there is a portion of time where csv is
outdated until the bump is merged, and it is generating RP and RPAs
with names like `136-..-1350-...`.


pkg/prowgen/prowgen_git.go
```diff
@@ -114,7 +114,7 @@ func CmpBranches(a string, b string) int {
 	return av.Compare(*bv)
 }
 
-func semverFromReleaseBranch(b string) (*semver.Version, error) {
+func SemverFromReleaseBranch(b string) (*semver.Version, error) {
```
creydr, line 117: just a nit: maybe we can move this into the pkg/soversion package

creydr (commented):
Thanks for fixing

/lgtm

PR #530, opened 2025-01-27
Run EC tests on push

pierDipi:
EC tests on PRs is constantly failing and it doesn't make sense as it's now just noise. We can re-evaluate if running it on PRs makes sense in the future but now it's just not working.


creydr (commented):
/lgtm

PR #531, opened 2025-01-27
Sort dependabot configurations by branch to be idempotent

pierDipi:
the configuration is getting updated just because we iterate on a map which doesn't guarantee ordering


creydr (commented):
Thanks for fixing
/lgtm

PR #536, opened 2025-01-28
Run make generate-release before running konfluxgen

pierDipi:
We're running `make generate-release` (or equivalent) after running konfluxgen. The issue with that is that the dockerfile generator is also generating the `rpms` lock file which is necessary to enable the prefetch dependencies task for rpms.

If we run it before, the RPMs lock file would be present.


pierDipi:
```
2025/01/28 13:44:42 Failed to generate Konflux configurations: %w failed to Generate dockerfiles with "make generate-release" for "openshift-knative/backstage-plugins" [release-v1.15]: [openshift-knative/backstage-plugins] failed to run make generate-release []: exec: "make generate-release": executable file not found in $PATH - 
exit status 1
```

creydr (commented):
/lgtm

PR #544, opened 2025-01-31
Remove nudges for SO components as we don't need them

pierDipi:
Nudges is just a mechanism to get auto updates of images that depends on other components, however, we're not really using them and they are causing a huge number of unnecessary PRs.

- https://github.com/openshift-knative/serverless-operator/issues?q=is%3Apr%20author%3Aapp%2Fred-hat-konflux%20%20%20
- https://github.com/openshift-knative/serverless-operator/issues?q=is%3Apr%20is%3Aopen%20author%3Aapp%2Fred-hat-konflux-kflux-prd-rh02%20

Ref: https://konflux-ci.dev/docs/how-tos/configuring/component-nudges

### DIFF
--- a/pkg/prowgen/prowgen_konflux.go
+++ b/pkg/prowgen/prowgen_konflux.go
@@ -25,8 +25,6 @@ import (
 
 const (
 	KonfluxBranchPrefix = "sync-konflux-"
-
-	NudgeFilesAnnotationName = "build.appstudio.openshift.io/build-nudge-files"
 )
 
 var hackRepo = Repository{Org: "openshift-knative", Repo: "hack"}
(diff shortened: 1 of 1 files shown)


pkg/prowgen/prowgen_konflux.go
```diff
@@ -412,12 +410,6 @@ func GenerateKonfluxServerlessOperator(ctx context.Context, openshiftRelease Rep
 			GlobalResourcesOutputPath:     resourceOutputPath,
 			PipelinesOutputPath:           fmt.Sprintf("%s/.tekton", r.RepositoryDirectory()),
 			Nudges:                        b.Konflux.Nudges,
```
creydr, line 412: The components also have the `build-nudges-ref`:
https://github.com/openshift-knative/hack/blob/71f8c5394687bc133ab7c0838a6b1364661700af/pkg/konfluxgen/dockerfile-component.template.yaml#L12

pierDipi: This stays there

pierDipi: The reason being, we might want to use them for some components (like client)

creydr: :+1: 
/lgtm

PR #546, opened 2025-01-31
Fix commit message for dependabot PR update

creydr:
Uses executed command in commit message, so it's a bit clearer and we don't have something like https://github.com/openshift-knative/serverless-operator/pull/3330/files#diff-f3e87f564ac7fc73df0a993a3baab40cbf2ba0c3023bb05ff4d3be55df5d04b1R42


pierDipi (approved):
/lgtm
/approve

pierDipi:
TIL about the syntax

PR #558, opened 2025-02-03
Exclude enterprise contract runs from dashboard by default

creydr:
For a better overview on the builds, we can filter the EC runs from the default view (can still be brought back, by adjusting the search query)


pierDipi (approved):
/lgtm
/approve

PR #559, opened 2025-02-04
Optimize dependabot updates with groups

pierDipi:
Group updates based on type (major, minor, patch) and security.

See docs:
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference


creydr (commented):
/lgtm

PR #560, opened 2025-02-04
Add exclude pattens to Dependabot configurations for groups

pierDipi:
`Ignores` are ignored :)

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--


creydr (approved):
/lgtm

PR #562, opened 2025-02-04
Create workflow directory before writing dependabot workflow and exclude istio

pierDipi:
As per title and Slack discussion


pkg/dependabotgen/dependabotgen.go
```diff
@@ -241,6 +246,10 @@ func WriteDependabotWorkflow(repoDir string, run string) error {
 		return nil
 	}
 
+	if err := os.MkdirAll(filepath.Join(repoDir, ghDir, workflowsDir), 0755); err != nil {
```
skonto, line 249: Was this working previously?

PR #566, opened 2025-02-10

serverless-qe:


pierDipi (approved):
/lgtm
/approve

config/backstage-plugins.yaml
```diff
@@ -1,10 +1,5 @@
 config:
   branches:
-    release-v1.12:
-      openShiftVersions:
-      - version: "4.15"
-      - onDemand: true
-        version: "4.12"
```
creydr, line 7: Somehow this got only removed in the backstage config. Other repos still have it:
https://github.com/openshift-knative/hack/blob/192cf872bf2af9e29eef139222d78b1057aada4e/config/eventing.yaml#L9-L14

PR #570, opened 2025-02-12
Skip presubmit OCP CI jobs when only Konflux configs change

pierDipi:
We don't need to run expensive tests when only Konflux config change


creydr (approved):
/lgtm

PR #574, opened 2025-02-12
Update skip if only changed for Konflux configs

pierDipi:
There is a misconfiguration https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/61453/pull-ci-openshift-release-master-config/1889595034622234624

```
is set to always run but also declares skip_if_only_changed targets
```


creydr (approved):
/lgtm

PR #579, opened 2025-02-13
fix image override naming for kn-plugin-func

gauron99:


gauron99:
@creydr ^^ PTAL :pray:

PR #580, opened 2025-02-13
:broom: Allow dockerfilegen reuse

cardil:
## Summary

This PR will allow reusing the dockerfilegen in other packages, for example from https://github.com/openshift-knative/deviate

/king cleanup

## Changes

- Extracting the logic of the generator to `./pkg/dockerfilegen` package for reuse
- Replaced `log.Fatal` with return of errors
- A bit more convenient defaults

## Testing

Tested versus kn-plugin-event and eventing-kafka-broker by running:

```shell
$ go run ./cmd/generate --root-dir ~/git/knative-eventing-kafka-broker
```


pkg/util/errors/wrap.go
```diff
@@ -0,0 +1,94 @@
+package errors
+
+import (
+	"fmt"
+	"io"
+)
+
+// Rewrap the error to make sure the errors support the Format function, to
+// dump the details. In Golang SDK the fmt.Errorf returns errors which do not
+// support the Format, especially with details `%+v`, so the details are hidden.
```
creydr, line 10: Not sure why `Rewarp` is needed and what you mean with `In Golang SDK the fmt.Errorf returns errors which do not support the Format, especially with details `%+v`, so the details are hidden`

pierDipi: I have the same concern, do we _really_ need it or can we live without given the usage of the tools?

cardil: In Go, the errors are required to implement the `Error() string` func. But errors could be complex objects, holding additional data like for ex.: Stacktrace. So, it is a good idea to return the error object, instead of stringified representation, for error handling with things like `errors.Is` and `errors.As`. Errors also could be wrapped - the built-in `fmt.Errorf()` does that.

To print all the details of the errors with details, you could use `%+v` formatter, [as used in many libraries](https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L168-L172).

However, [the Golang built-in wrap errors](https://github.com/golang/go/blob/1eb4c0dcb125d27e1a4296ae136f75ac08c3b9c5/src/fmt/errors.go#L54-L78) don't respect the `+` as other libraries do. This function fixes that, by implementing proper format function for wrapped errors.

pierDipi (approved):
/lgtm
/approve

PR #582, opened 2025-02-13
:broom: Skipping source dockerfile for vendorless projects

cardil:
Just skipping the source Dockerfile if it is not needed.

/kind cleanup

Part of [SRVCLI-405](https://issues.redhat.com/browse/SRVCLI-405)

### DIFF
--- a/pkg/dockerfilegen/generator.go
+++ b/pkg/dockerfilegen/generator.go
@@ -182,8 +182,12 @@ func generateDockerfile(params Params, mainPackagesPaths sets.String) error {
 		return err
 	}
 
-	if _, err = saveDockerfile(d, DockerfileSourceImageTemplate, params.Output, params.DockerfilesSourceDir); err != nil {
-		return err
+	if dirExists(path.Join(params.RootDir, "vendor")) {
+		if _, err = saveDockerfile(d, DockerfileSourceImageTemplate, params.Output, params.DockerfilesSourceDir); err != nil {
+			return err
+		}
+	} else {
+		log.Println("Skipping source dockerfile for vendorless projects")
 	}
 
 	var additionalInstructions []string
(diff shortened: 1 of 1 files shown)


pkg/dockerfilegen/generator.go
```diff
@@ -182,8 +182,12 @@ func generateDockerfile(params Params, mainPackagesPaths sets.String) error {
...
 
-	if _, err = saveDockerfile(d, DockerfileSourceImageTemplate, params.Output, params.DockerfilesSourceDir); err != nil {
-		return err
+	if dirExists(path.Join(params.RootDir, "vendor")) {
+		if _, err = saveDockerfile(d, DockerfileSourceImageTemplate, params.Output, params.DockerfilesSourceDir); err != nil {
+			return err
+		}
+	} else {
+		log.Println("Skipping source dockerfile for vendorless projects")
 	}
```
pierDipi, line 191: I don't think having vendor vs not is the reason it needs to have a source image, the main reason is that it might be needed to be aggregated for SO to run tests on the full operator: https://github.com/openshift-knative/serverless-operator/blob/main/openshift/ci-operator/source-image/Dockerfile

cardil: Hmm, only [specific sources are used in S-O](https://github.com/openshift-knative/serverless-operator/blob/ab8cf342a99f99162a976bfa92333f9f1b78bd5f/openshift/ci-operator/source-image/Dockerfile#L8-L11). Also, the source-image template has only those vendor related instructions: https://github.com/openshift-knative/hack/blob/cb8a7c327cd11b4554bf2cc813ed99966e55a012/pkg/dockerfilegen/dockerfile-templates/SourceImage.dockerfile.tmpl#L5-L7

What's more, the `generate-groups.sh` and `generate-internal-groups.sh` are [already defunct](https://github.com/openshift-knative/serving/blob/6ce6512d82389023c03c134e2fb68c9d38c6fd75/vendor/k8s.io/code-generator/generate-groups.sh#L17-L20) in release-1.17, and the above instructions don't make sense anymore. Only the `generate-knative.sh` is still there.

But, I get you. Having this source images doesn't hurt, much.

cardil: /hold

PR #585, opened 2025-02-14
:gift: Scan for main packages on imports

cardil:
For vendorless projects we need to collect the test images from deps. This does that by scanning imports (usually the `hack/tools.go` file).

Part of [SRVCLI-405](https://issues.redhat.com/browse/SRVCLI-405)

/kind enhancement


pierDipi:
As discussed on Slack, for `kn-event-plugin` we can leverage the reconciler-test file resolver https://github.com/knative-extensions/reconciler-test/tree/main/pkg/images/file with the help of the generate command as used here https://github.com/openshift-knative/eventing-istio/blob/e8680a4b989140e96986488dfbd76d1028f4a557/openshift/generate.sh#L48-L49

The benefits are:

- faster
- less resources
- less builds (big benefits as Konflux scalability is still a big challenge)
- more environmental friendly
- Cost savings
- less code to maintain as we wouldn't need scanning
- Reuse an existing feature

In the meantime, we decided that:
- kn-event-plugin will generate the production images with generate
- test images would be configured with static-images (custom Dockerfiles), similar to how Java images are maintained in EKB https://github.com/openshift-knative/eventing-kafka-broker/tree/main/openshift/ci-operator/static-images/dispatcher
- For 1.37+, migrate to file resolver

PR #586, opened 2025-02-17
:gift_heart: Configure dockerfilegen flags via struct fields

cardil:
This allows to configure the flags using struct fields. This allows to read the parameters from either the command line or the config file, say a YAML.

/kind enhancement


pierDipi (approved):
/lgtm
/approve

PR #591, opened 2025-02-19
MintMaker: Enable tekton manager only

pierDipi:
We use dependabot for other deps and we get better grouping of updates


creydr (approved):
/lgtm

PR #597, opened 2025-02-24
Push renovate.json to default branch along with dependabot

pierDipi:
Renovate reads only the default branch, similar to dependabot

See https://redhat-internal.slack.com/archives/C04PZ7H0VA8/p1740071050329519?thread_ts=1739798087.812689&cid=C04PZ7H0VA8


cardil (approved):
/lgtm

PR #601, opened 2025-03-03
Use OCP image based on Golang version

creydr:
Setting the OCP image version based on the used Golang version


pierDipi (approved):
/lgtm
/approve

PR #598, opened 2025-02-26
Update rpms lock for tzdata and rsync

Kaustubh-pande:
update `tzdata to tzdata-2025a-1.el8.noarch` and update `rsync to rsync-3.1.3-20` to fix failures at https://github.com/openshift-knative/eventing/pull/1154

cc @pierDipi

PR #606, opened 2025-03-06
Auto update RPMs lock files

pierDipi:
This will auto fix issues like https://github.com/openshift-knative/hack/pull/598 automatically


creydr (approved):
/lgtm

PR #609, opened 2025-03-06
Bump to go 1.24

creydr:
Bumping Golang, as we have some midstream repos with 1.23 now (bumping here directly to 1.24)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift-knative/hack
 
-go 1.22.9
+go 1.24.0
 
 require (
 	github.com/blang/semver/v4 v4.0.0
--- a/pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
+++ b/pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
@@ -3,7 +3,7 @@
 FROM registry.ci.openshift.org/ocp/4.17:cli-artifacts as tools
 
 # Dockerfile to bootstrap build and test in openshift-ci
-FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 as builder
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift-knative/hack
 
-go 1.22.9
+go 1.24.0
```
dsimansk, line 3: Are we driving default Golang version by inferring from hack's `go.mod`? In addition doesn't this min version break `go run/go install` we use everywhere to run various tools from o-k/hack?

pkg/project/testoutput/openshift/ci-operator/build-image/Dockerfile
```diff
@@ -3,7 +3,7 @@
 FROM registry.ci.openshift.org/ocp/4.17:cli-artifacts as tools
 
 # Dockerfile to bootstrap build and test in openshift-ci
-FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.22-openshift-4.17 as builder
+FROM registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.24-openshift-4.19 as builder
```
dsimansk, line 6: This image doesn't exist in `registry.ci`. Only `-1.23-openshift-4.19` combination.

creydr: This shouldn't be a big issue here I guess, as this is only used for unit tests. But good to know

go.mod
dsimansk: In cases where the infra is on 1.23 still.

dsimansk: Let's use a common go setup then. And it needs bump to 1.23 as well. 

https://github.com/openshift-knative/hack/blob/ba4c9923390aec1ef4f47093af5bab8e8861bd4a/actions/setup-go/action.yaml

creydr: Closing in favor of https://github.com/openshift-knative/hack/pull/610 then
/close

PR #610, opened 2025-03-06
Bump default golang version to 1.23 in setup-go action

creydr:
Bumping the default golang version to 1.23 in setup-go action and use this action in the Konflux related workflows

### DIFF
--- a/.github/workflows/apply-konflux-manifests.yaml
+++ b/.github/workflows/apply-konflux-manifests.yaml
@@ -20,9 +20,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Golang
-        uses: actions/setup-go@v5
-        with:
-          go-version-file: go.mod
+        uses: ./actions/setup-go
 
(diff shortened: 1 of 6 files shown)


.github/workflows/apply-konflux-manifests.yaml
```diff
@@ -20,9 +20,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Golang
-        uses: actions/setup-go@v5
-        with:
-          go-version-file: go.mod
+        uses: ./actions/setup-go
```
dsimansk, line 23: is this path ok? in comparison to the rest

creydr: As we checkout the repo into the root folder, I'd think so

creydr: In the other ones, we checkout the hack repo to a custom path

dsimansk (approved):
/approve
/lgtm

PR #611, opened 2025-03-07
Skip tests when only '^rpms.lock.yaml$|^hack/.*|^OWNERS.*|.*\\.md' change

pierDipi:
Further optimizations for expensive tests. No need to run them when only:
- rpms.lock.yaml 
- hack 
- owners
- docs

change

### DIFF
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -394,7 +394,7 @@ func AlwaysRunInjector() JobConfigInjector {
 					// Prevent sneaking in wrong settings from previous runs of "make jobs".
 					// Make sure we reset it to default.
 					jobConfig.PresubmitsStatic[k][i].AlwaysRun = false
-					jobConfig.PresubmitsStatic[k][i].SkipIfOnlyChanged = "^.tekton/.*|^.konflux.*|^.github/.*"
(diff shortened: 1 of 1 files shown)


pkg/prowgen/prowgen.go
```diff
@@ -394,7 +394,7 @@ func AlwaysRunInjector() JobConfigInjector {
 					// Prevent sneaking in wrong settings from previous runs of "make jobs".
 					// Make sure we reset it to default.
 					jobConfig.PresubmitsStatic[k][i].AlwaysRun = false
-					jobConfig.PresubmitsStatic[k][i].SkipIfOnlyChanged = "^.tekton/.*|^.konflux.*|^.github/.*"
+					jobConfig.PresubmitsStatic[k][i].SkipIfOnlyChanged = "^.tekton/.*|^.konflux.*|^.github/.*|^rpms.lock.yaml$|^hack/.*|^OWNERS.*|.*\\.md"
```
creydr, line 397: maybe include the `.konflux-release` folder from SO with the override-snapshots too

creydr (approved):
/lgtm

/hold 
in case you want to address the comment too

pkg/prowgen/prowgen.go
pierDipi: yes, it should covered

PR #626, opened 2025-03-13
Konflux Dashboard: Filter out test builds by default

creydr:
Don't show builds for tests by default in the Konflux dashboard


pierDipi (approved):
/lgtm
/approve

PR #638, opened 2025-03-26
Generate the rpms.lock file, as soon as we have additional packages

creydr:
We should generate the rpms.lock file, as soon as we have additional packages to be installed


pierDipi (approved):
/lgtm
/approve

PR #640, opened 2025-03-26
Add Dependabot for backstage-plugins NPM

pierDipi:
As per title, each dir points to a dir with a `package.json`


matzew (approved):
/lgtm

PR #644, opened 2025-04-04
Update rpm lockfile prototype ref to v0.15.0

creydr:
Updating ref to v0.15.0 tag to have updates from https://github.com/konflux-ci/rpm-lockfile-prototype/pull/90/files

This should help on issues like https://github.com/openshift-knative/hack/actions/runs/14239958985/job/39922767359

### DIFF
--- a/.github/workflows/release-discover-branches.yaml
+++ b/.github/workflows/release-discover-branches.yaml
@@ -53,7 +53,7 @@ jobs:
         with:
           repository: 'konflux-ci/rpm-lockfile-prototype'
           path: ./src/github.com/openshift-knative/rpm-lockfile-prototype
-          ref: '6d222512f5e7b89c59fdcf9bddc9f4d535f3f255'
+          ref: 'v0.15.0'
 
       - name: Install podman
         run: |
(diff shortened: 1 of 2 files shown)


creydr:
/hold
still not working

.github/workflows/release-discover-branches.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
         with:
           repository: 'konflux-ci/rpm-lockfile-prototype'
           path: ./src/github.com/openshift-knative/rpm-lockfile-prototype
-          ref: '283ee2cd7938a2142d8ac98de33ba0d0e3ac146f'
+          ref: '6d222512f5e7b89c59fdcf9bddc9f4d535f3f255'
```
Kaustubh-pande, line 56: Maybe we can change it to ref: `main`

creydr:
https://github.com/konflux-ci/rpm-lockfile-prototype/pull/90/files seem to fix it

.github/workflows/release-discover-branches.yaml
creydr: Updated to tag v0.15.0 which has the fix. This should also help us to keep the build reproduceable until it is more stable

creydr:
@rudyredhat1 @Kaustubh-pande can you recheck? The discover branches job succeeded

PR #669, opened 2025-04-23
Don't generate branch protection for SO

creydr:
SO does not have branch protection rules for release-next and release-next-ci. So we don't set those for SO

https://github.com/openshift/release/blob/363307d181d1cf4734191bb794be20df54431b7d/core-services/prow/02_config/openshift-knative/serverless-operator/_prowconfig.yaml


cardil (approved):
/lgtm

PR #674, opened 2025-04-23
:gift: Use 4.18 cluster pool

cardil:
Use 4.18 cluster pool across the whole Serverless


cardil:
The cluster pool is setup here: https://github.com/openshift/release/pull/64192 and it is sitting in standby state in the table: https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools so everything should be ready for the switch

creydr (approved):
/lgtm

PR #685, opened 2025-05-05
Add target branch name to renovate PR name

creydr:
This adds the target name as a prefix to the name of the PR from renovate. This affects the `Update Konflux references` PRs and results in something like `[release-v1.15] Update Konflux references` and makes it thus clearer from the overview which branch a PR targets.

Tested for example in https://github.com/openshift-knative/eventing/pull/1368


rudyredhat1:
thank you, it will help a lot
/lgtm
/approve

PR #686, opened 2025-05-06
Fix renovate template to use correct delimiters for vars

creydr:
Should avoid `failed to parse renovate template: template: renovate.template.json:4: function "baseBranch" not defined` like in https://github.com/openshift-knative/hack/actions/runs/14852438606/job/41698417657


creydr:
@rudyredhat1 @Kaustubh-pande can you PTAL?

PR #695, opened 2025-05-19
update fbc policy for EC and stage push failure

rudyredhat1:
Hi @creydr 
As per the slack thread discussion, the stage push failure we have this workaround.
slack thread - 1 = https://redhat-internal.slack.com/archives/CKR568L8G/p1747375927922199
slack thread - 2 = https://redhat-internal.slack.com/archives/C08A979HE48/p1747647489361809

Can you PTAL?
Thanks


creydr (commented):
Looking good. But we need to wait with this, until we have the policy setup in Gitlab and the name confirmed

/hold

PR #709, opened 2025-05-27
Add FBC Release CRs from release-1.36 revision for stage

serverless-qe:
Add FBC Release CRs from release-1.36 revision for stage


Kaustubh-pande:
FBC pipeline passed won't required this PR

PR #736, opened 2025-06-18
Add component Release CRs from release-1.36 revision for prod

serverless-qe:
Add component Release CRs from release-1.36 revision for prod


bj2122:
lgtm

PR #684, opened 2025-04-30
Add new midstream branches and update pipelines

serverless-qe:
Discover new branches and update Konflux pipelines using openshift-knative/hack.

PR #740, opened 2025-06-26
migrate apply-tag from 0.1 to 0.2

Kaustubh-pande:
as per Migration from 0.1 to 0.2 [see](https://github.com/openshift-knative/serverless-operator/pull/3667)

ref: https://github.com/konflux-ci/build-definitions/blob/main/task/apply-tags/0.2/MIGRATION.md

cc @rudyredhat1


creydr:
@Kaustubh-pande should this come through https://github.com/openshift-knative/hack/pull/684/files#diff-57f308fdab736a2c329e454b964ba038e8ea1adf3a7a542d773683c01236d511R235 ?

PR #744, opened 2025-07-15
update go version tov1.23.10

Kaustubh-pande:
as per title
cc @creydr @rudyredhat1


Kaustubh-pande:
The previous Go version error was solved, but the test fails this time.

PR #747, opened 2025-07-16
upgrade to  OCP 4.19, use 4.20 as a candidate

maschmid:


creydr:
> /hold until we rehearse the updated jobs in openshift/release

You have have a DNM PR in openshift/release for it open?

PR #750, opened 2025-07-16
Add sync for OWNERS files

creydr:
Lets us define reviewers and approvers in the repo config in hack and distribute it to the repos and all maintained branches. 
For now the changes are distributed via the "Sync Konflux configurations" PRs. We could also do this via a separate PR, but I didn't want to increase the noise.

I took the current owners config from the repos main branches and removed users which are not in the team anymore.


cardil (approved):
/lgtm

PR #753, opened 2025-07-16
downgrade go 1.23.10 to 1.23.9

maschmid:
as that is the version registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.23-openshift-4.19 currently has


maschmid:
superseded by https://github.com/openshift-knative/hack/pull/754

PR #762, opened 2025-07-22
Move `skip_if_only_changed` to config folder

creydr:
We cannot change the `always_run` prow configs under ci-operator/jobs anymore (see [Slack](https://redhat-internal.slack.com/archives/CBN38N3MW/p1752996614456229), [Slack](https://redhat-internal.slack.com/archives/CBN38N3MW/p1752835745451229) and https://github.com/openshift/release/pull/67251/files). So moving the `skip_if_only_changed` configuration to the normal test setup in ci-operator/configs.
To update still set the `always_run` field on image builds, we are leveraging the exception mentioned [here](https://redhat-internal.slack.com/archives/CBN38N3MW/p1752996614456229), so not run all image builds on all PRs. This is a known hack for image builds and helps until we have support for `skip_if_only_changed` on image builds.

The diff of the testrun can be seen here for now https://github.com/openshift/release/pull/67356/files


creydr:
/hold
Something is invalid on the config

PR #763, opened 2025-07-24
re-enable prowgen for supported SO branches

maschmid:
- re-enable prowgen for supported SO branches
- drop onDemand from our SO main on 4.19 (our target for continuous jobs)
- try to fix an issue that caused generation of both RunIfChanged and SkipIfOnlyChanged  (for soak and ui tests, which declare `runIfChanged` explicitly


creydr (approved):
/lgtm

PR #764, opened 2025-07-29
Use RHEL 8 variant for Golang 1.24 builder image

creydr:
RHEL 8 variant for Golang 1.24 seems to be available nevertheless.

```
$ docker pull registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.24-openshift-4.20                                                                           
rhel-8-release-golang-1.24-openshift-4.20: Pulling from openshift/release
c7965aa70860: Downloading [==========>                                        ]  16.14MB/78.96MB
45b08c48498c: Downloading [>                                                  ]  1.081MB/697.1MB
8a9007f9a0f1: Downloading [>                                                  ]  540.7kB/110.9MB
1f3c9d2ed5a8: Waiting 
^C%                                                                                                                                                                                                                                                                                                                                                           

$ docker pull brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_8_golang_1.24                                                                                                                                                                                                                                                            
rhel_8_golang_1.24: Pulling from rh-osbs/openshift-golang-builder
c7965aa70860: Pulling fs layer 
e69c4d9b35de: Pulling fs layer 
^C%   
```

PR #765, opened 2025-07-29
Use Golang 1.24 for SO >= 1.38 (and components >= 1.18) in Konflux

creydr:
Use Golang 1.24 RHEL 8 for SO >= 1.38 (and thus components >= 1.18).

#764 is to have the same in Prow


gauron99 (commented):
/lgtm
/approve
/hold unhold when you please

PR #770, opened 2025-07-31
Update kn plugins owners file

dsimansk:
Adding @maschmid and @Kaustubh-pande to kn's plugin repos to make sure we have a critical approvers to move either component's PR or CI configs. (similarly to Serving components in previous PR)

/cc @creydr @maschmid


creydr (approved):
/lgtm

PR #772, opened 2025-08-07
Bump to Golang 1.24 as default in shared setup-go action

creydr:
As per title


creydr:
@rudyredhat1 @Kaustubh-pande can you PTAL? This should help to fix the current issues on generate-ci (https://github.com/openshift-knative/hack/actions/runs/16798860761)

PR #790, opened 2025-10-03
Add support for RHEL 9 base images

dsimansk:
/cc @Kaustubh-pande @rudyredhat1 @creydr

### DIFF
--- a/.github/workflows/release-generate-ci.yaml
+++ b/.github/workflows/release-generate-ci.yaml
@@ -125,7 +125,7 @@ jobs:
           repository: openshift-knative/serving
           token: ${{ secrets.SERVERLESS_QE_ROBOT }}
       - name: Freeup disk space
-      - run: |
+        run: |
           sudo rm -rf "/usr/local/share/boost" || true
           sudo rm -rf /usr/share/dotnet || true
           sudo rm -rf /usr/local/lib/android || true
--- a/.github/workflows/release-generate-ci.yaml
+++ b/.github/workflows/release-generate-ci.yaml
@@ -20,10 +20,6 @@ jobs:
     name: generate-ci
     runs-on: ubuntu-latest
     steps:
-      - name: Freeup disk space
-        run: |
-          sudo rm -rf "/usr/local/share/boost"
-          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
       - if: ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule') && github.ref_name == 'main' }}
         name: '[backstage-plugins] Clone repository'
         uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


pkg/dockerfilegen/generator.go
```diff
@@ -152,9 +152,27 @@ func generateDockerfile(params Params, mainPackagesPaths sets.Set[string]) error
...
+		if !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("%w: Failed to read project metadata file: %w",
+				ErrBadConf, errors.WithStack(err))
+		}
+		log.Println("File not found:", params.ProjectFilePath, "(Using defaults)")
+		metadata = project.DefaultMetadata()
+	}
+
+	rhelVersion := "rhel-8"
+	minorVersion, err := strconv.Atoi(strings.Replace(metadata.Project.Tag, "knative-v1.", "", 1))
```
Kaustubh-pande, line 166: Is metadata.Project.Tag coming from (SO) project.yaml? I don't see it project.yaml

pkg/dockerfilegen/generator.go
```diff
@@ -152,9 +152,27 @@ func generateDockerfile(params Params, mainPackagesPaths sets.Set[string]) error
...
+	if err != nil {
+		if minorVersion >= 17 {
```
Kaustubh-pande, line 168: Or might be we can compare with project.version >="1.37.0"?

pkg/dockerfilegen/generator.go
```diff
@@ -152,9 +152,35 @@ func generateDockerfile(params Params, mainPackagesPaths sets.Set[string]) error
...
+		// tag before knative-v1.17
+		minorVersion, err := strconv.Atoi(strings.Replace(metadata.Project.Tag, "knative-v1.", "", 1))
+		if err != nil {
+			if minorVersion < 17 {
+				rhelVersion = "rhel-8"
+			}
+		}
+	} else {
+		// version before 1.37+
+		if metadata.Project.Version == "1.36.1" || metadata.Project.Version == "1.35.1" {
```
Kaustubh-pande, line 176: Maybe we can use just the safer side if we do a patch release 1.36.2 or newer

`minor, _ := strconv.Atoi(strings.Split(metadata.Project.Version, ".")[1])
if minor <= 36 {
	rhelVersion = "rhel-8"
} else {
	rhelVersion = "rhel-9"
}
`

dsimansk:
@Kaustubh-pande @creydr I've added disk free up commands to template and CI action. It should be good for review now.

.github/workflows/release-generate-ci.yaml
```diff
@@ -230,246 +237,6 @@ jobs:
           fi
           gh pr create --base "$target_branch" --head "serverless-qe:$branch" --title "[$target_branch] Update dependabot configurations" --body "Update dependabot configurations" || true
         working-directory: ./src/github.com/openshift-knative/hack/openshift-knative/backstage-plugins
-      - env:
-          GH_TOKEN: ${{ secrets.SERVERLESS_QE_ROBOT }}
```
creydr, line 234: why is all of this removed? I think this was only for testing
btw: You can run `make generate-ci-action` to update this file too

dsimansk: should be fixed now

PR #791, opened 2025-10-07
Add CPE label and RHEL version placeholder in Dockerfile generator

Kaustubh-pande:
- Introduce CPE label support for images
- Add placeholder for RHEL version to dynamically set OS version in Dockerfiles

cc @dsimansk

### DIFF
--- a/pkg/dockerfilegen/templates.go
+++ b/pkg/dockerfilegen/templates.go
@@ -2,24 +2,19 @@ package dockerfilegen
 
 import "embed"
 
-//go:embed dockerfile-templates/Default.dockerfile.tmpl
-//go:embed dockerfile-templates/rhel-9/Default.dockerfile.tmpl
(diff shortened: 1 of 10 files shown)


dsimansk:
This place needs some attention: https://github.com/openshift-knative/hack/blob/main/pkg/dockerfilegen/generator.go#L281-L294

pkg/dockerfilegen/templates.go
```diff
@@ -2,19 +2,19 @@ package dockerfilegen
 
 import "embed"
 
-//go:embed dockerfile-templates/Default.dockerfile.tmpl
+//go:embed dockerfile-templates/Default.dockerfile.tmpl dockerfile-templates/rhel-9/*.tmpl
```
dsimansk, line 5:
```suggestion
//go:embed dockerfile-templates/Default.dockerfile.tmpl dockerfile-templates/rhel-9/*
```

It should be dir input, otherwise pattern match with suffix doesn't work.

PR #917, opened 2026-03-06
Update direct golang deps

dsimansk:
### Changes:
- a bit of direct dependencies bumps
- in addition refactor of overlapping library use or move to golang's stdlib 

/cc @creydr

PR #947, opened 2026-04-10
Cleanup injection of image build configs leftovers

dsimansk:
A follow-up to change merged in #917.

/cc @maschmid @creydr


maschmid:
/lgtm

looks good to me twice

PR #953, opened 2026-04-14
increase size of the vsphere cluster to 3 nodes

maschmid:
We see EKB dispatchers not able to scale up due to insufficient resources on the cluster, e.g. https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-knative-serverless-operator-main-420-vsphere-e2e-vsphere-continuous/2041734414412550144

increasing the worker node count from the default 2 to 3 seems to help:

tested on https://github.com/openshift/release/pull/77540


dsimansk (approved):
/approve
/lgtm

PR #954, opened 2026-04-14
Refactor slack notification to use .prowgen config file

dsimansk:
Changes: 
 - Refactor slack notification away from job injector to .prowgen file
   - `.prowgen` file will be [refactored](https://github.com/openshift/ci-tools/pull/5012) into `ReleaseBuildConfiguration` struct, but from our point of view it will be straight-forward refactor nesting the structs together 
   - Current set of changes removes decorator pattern and eliminates double exec of `make jobs` on the openshift/release repo 
     - as a result generator pipeline run should be a bit quicker as a result 🤞  

Preview of changes: https://github.com/openshift/release/pull/77640

/cc @maschmid @creydr @Kaustubh-pande

### DIFF
--- a/config/backstage-plugins.yaml
+++ b/config/backstage-plugins.yaml
@@ -52,9 +52,6 @@ repositories:
 - dockerfiles: {}
   e2e:
   - match: .*e2e-tests$
-  ignoreConfigs:
-    matches:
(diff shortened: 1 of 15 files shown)


dsimansk:
For CI 
/hold

PR #959, opened 2026-04-16
update OCP versions for SO 1.37 and 1.38

maschmid:
Only touching the SO 1.37/1.38 (knative 1.17/1.21) configurations.

Make 4.22 the candidate OCP, use 4.21 as the default.
For SO 1.38, update the older OCP version from 4.14 to 4.16
Add 4.22 candidate on-demand jobs where 1.17 had them, but 1.21 didn't .

### DIFF
--- a/config/backstage-plugins.yaml
+++ b/config/backstage-plugins.yaml
@@ -29,9 +29,9 @@ config:
       - candidateRelease: true
         onDemand: true
         skipCron: true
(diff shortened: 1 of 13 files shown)


maschmid:
DNM opened here,  https://github.com/openshift/release/pull/77905

dsimansk:
@maschmid thanks a lot!

PR #964, opened 2026-04-22
Add override-snapshot generation to CRS pipeline

dsimansk:
On the S-O side we can drop periodic genration of `override-snapshot.yaml` file.

It's only needed whenever being applied, so we can just generated it mid-flight. Wdyt?
 

/cc @creydr @Kaustubh-pande


Kaustubh-pande:
Good call, it will also reduce a lot of unnecessary snapshot updates.
/lgtm
/approve

feel free to unhold this.

creydr:
I think one reason why we generate it the current way, was to have them persisted in Git and thus see directly which component versions a release has (as background information)

PR #972, opened 2026-05-20
Enable automatic Go toolchain upgrades in release workflow

Kaustubh-pande:
## Summary

Enable `GOTOOLCHAIN=auto` for the release generation workflow.

`actions/setup-go` installs the Go version defined in `go.mod`, but some dependencies from `openshift-knative/hack` may require a newer patch version than the one currently specified.

Without automatic toolchain upgrades enabled, the workflow fails with errors similar to:

```text
requires go >= 1.25.5 (running go 1.25.0; GOTOOLCHAIN=local)
```

Setting `GOTOOLCHAIN=auto` allows Go to automatically download and use a newer compatible toolchain version when required, while still defaulting to the version defined in `go.mod`.

This also makes the workflow more future-proof for future patch version updates.

### DIFF
--- a/.github/workflows/run-command-repository.yaml
+++ b/.github/workflows/run-command-repository.yaml
@@ -33,7 +33,6 @@ jobs:
         uses: actions/setup-go@v6
         with:
           cache: false
-          go-version-file: "./src/github.com/${{ inputs.repository }}/go.mod"
 
       - name: Install prerequisites
         env:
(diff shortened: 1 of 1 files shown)


dsimansk:
Let's remove this config: https://github.com/openshift-knative/hack/pull/972/changes#diff-87d2cd64fd1c0016712d425a18825808506c28d04ccef2cc333a99941a90b50fR36

PR #978, opened 2026-06-02
Add FBC Release CRs from release-1.37 revision for stage

serverless-qe:
Add FBC Release CRs from release-1.37 revision for stage


Kaustubh-pande:
we will do it with https://github.com/openshift-knative/hack/pull/980

PR #995, opened 2026-07-09
skip cron for all except release-v1.17 (SO 1.37) and 1.21 (SO 1.38)

maschmid:
as only those two releases are interesting ATM


dsimansk:
@maschmid thanks, great idea!

PR #1000, opened 2026-08-12
use konflux index in lp jobs

maschmid:
to get rid of flaky on-cluster index s2i builds


maschmid:
Is it supposed to  work now? On the rehearse we see the 1.37.2 not coming up, I don't remember exactly why, but perhaps because at that time we didn't have an index with the 1.37.2 konflux build yet?

maschmid:
Actually, for the lp-interop tests we should probably just use the redhat index anyway and run with current Serverless there...   (I think we don't have an option for that yet in the scripts though

